PR #61, opened 2018-11-28
Add section on model security

Benjamin-Lee:
I think we should dedicate a section to talk about model security, specifically within the context of biology.


cgreene:
I agree that we should really talk about the hazards of sharing models trained on data that cannot be ethically shared. I am happy to contribute to the writing of this rule.

PR #65, opened 2018-11-29
Update contributors.md

smsaladi:
I wasn't sure if/when to add myself, but I guess that's now


agitter:
You're right, we've asked contributors to add themselves here as soon as they join in the discussion in an issue.

Issue #64, opened 2018-11-29
Moving forward (meta)


pstew:
* Have we identified corresponding author(s) to help keep this on track and importantly pay the ~$2k USD publication fee?
* Should we agree on some timeline for getting the rules finalized and the presubmission inquiry sent in? Should the corresponding author(s) facilitate this?
* Will @Benjamin-Lee submit the presubmission inquiry? 
* If the presubmission inquiry goes well, then we should form a plan about who works on what or do we just see how it turns out? (some discussion in #48 about this)
* If the presubmission inquiry does not go well, then do we try a different journal with the same format or turn this into a review/primer or maybe a Methods in Molecular Biology chapter?
* Do some of the authors communicate offline? Some of you seem to work/have worked together on other projects. If so, can these communications/decisions be shared with us?

agitter:
> Do some of the authors communicate offline? Some of you seem to work/have worked together on other projects. If so, can these communications/decisions be shared with us?

I advocate for keeping all the communications public on GitHub.  That keeps the project open and inclusive and also helps make everyone's contributions transparent.  At some point, journal communications will be made via email.  For the deep review, we posted most those as GitHub issues to keep all co-authors informed.

I personally do happen to know other contributors here from prior projects or affiliations but still plan to keep the manuscript discussion on GitHub.  Several of us know each other from the collaborative [deep review](https://github.com/greenelab/deep-review/).

rasbt:
> I'm currently leaning towards using GH projects to track each rule's development. That way, we can use issues to track things to work on for each rule and use assignments to make sure that people know who is working on what.

that's sounds like a good idea. 

> I have no experience publishing book chapters, so expert guidance here is also welcome.

There are probably exceptions, but these are usually by invitation afaik.

> review/primer or maybe a Methods in Molecular Biology chapter?

Based on my experience writing for them they really want to be hands-on and have actionable examples (e.g, in this case usage instructions). We could do something like that but would probably make some substantial change.

Maybe we can worry about these things after the pre-submission  and pre-submission response :)

agitter:
For deep review, we let authorship and the author roles evolve organically.  @cgreene was the initial corresponding author.  Over time, I helped more and he invited me to join as corresponding.  We didn't really plan it in advance.

[...]

PR #67, opened 2018-11-30
Create authors.tsv

Benjamin-Lee:
Per #64.

### DIFF
--- a/authors.tsv
+++ b/authors.tsv
@@ -1,0 +1,2 @@
+Name	Substantial contributions to the conception or design of the work; or the acquisition, analysis, or interpretation of data for the work	Drafting the work or revising it critically for important intellectual content	Final approval of the version to be published	Agreement to be accountable for all aspects of the work in ensuring that questions related to the accuracy or integrity of any part of the work are appropriately investigated and resolved
+Benjamin Lee	x		x	x


Benjamin-Lee:
As an example, I included myself. Since there is no content (yet), the `Drafting the work` column is blank. Once we start writing the paper, we can use this file to keep track of the authors who have met the ICMJE requirements. 

Should we convert contributors.md into this file?

rasbt:
I think I would convert the contributors file into that format. However, in the end, even if people don't meet the authorship requirements, we want to list them as contributors in an Acknowledgement section, right?

The question is are we planning to edit this file during working on the paper? I am bit worried that this could end up in an incentive where, once there is a checkmark, the contribution efforts might drop :P. So maybe it would be better to keep a file of what was contributed and then discuss authorship once we have a manuscript?

Benjamin-Lee:
> I am bit worried that this could end up in an incentive where, once there is a checkmark, the contribution efforts might drop

That's a really good point that I hadn't thought about. I suppose a contribution-ordered author list could help that, but you're right that once you're an author the incentive drops significantly.

agitter:
There is a possibility that once someone satisfies the minimum criteria for authorship they will drop out.  Author ordering or detailed contributions statements (or data) could help make sure everyone is given fair credit for their contributions.  For example, @dhimmel made very cool [contribution analyses ](https://greenelab.github.io/meta-review/#fig:contrib) of the deep review.  Manubot doesn't do that automatically, but his [code](https://github.com/greenelab/meta-review/tree/master/analyses/deep-review-contrib) could be adapted for this project.

However, any metric derived solely from commits and words edited provides only partial information.  Tracking GitHub discussions, pull request reviews, etc. as we proceed is helpful to complete the picture.

> even if people don't meet the authorship requirements, we still want to list them as contributors in an Acknowledgement section, right?

Yes, I'd say so.  If we merge `contributors.md` and `authors.tsv` we may want a new column for general discussion.  It isn't one of the ICMJE authorship criteria but would be the first column a new contributor checks when they add themselves.

For the tsv file itself, should we simplify the header row?
- `Discussion` (optional new column)
- `Substantial contributions`
- `Drafting`
- `Final approval`
- `Accountable`

dhimmel:
Is there a reason why the author information cannot go directly into `metadata.yaml`? If you didn't want the all contributors included as authors yet, you could make `contributors.yaml`. In the deep review, I think we found the YAML worked better than the TSV for author info.

Benjamin-Lee:
I propose moving authorship tracking to contributors.md so that there is once centralized place to track contributions that also renders nicely into HTML (unlike YAML).

agitter:
This tabular format is fine with me.  However, I suggest waiting to check the `Final approval` column until after the manuscript is drafted.

Benjamin-Lee:
@agitter done!

PR #76, opened 2018-12-07
Middle initial for me too

chevrm:


jmschrei:
Looks good to me too.

PR #81, opened 2018-12-09
Move capacity argument to overfitting section

rasbt:
I thought that moving the point 

  - Emphasize the added dangers of deep models with greater representational capacity

from "10. Don't share models trained on sensitive data." to "7. Address deep neural networks' increased tendency to overfit the dataset" makes sense, unless I am misunderstanding what was meant by that

### DIFF
--- a/rules.md
+++ b/rules.md
@@ -4,6 +4,7 @@
   - Garbage In/Garbage Out ([#26](https://github.com/Benjamin-Lee/deep-rules/issues/26))
   - Normalize your data ([#7](https://github.com/Benjamin-Lee/deep-rules/issues/7))
   - Make sure data is not biased/skewed ([#43](https://github.com/Benjamin-Lee/deep-rules/issues/43))
+  - How to deal with class imbalance
   - Make it reproducible ([#21](https://github.com/Benjamin-Lee/deep-rules/issues/21))
   - Validate it ([#27](https://github.com/Benjamin-Lee/deep-rules/issues/27))
   - Test/train/validate splitting ([#20](https://github.com/Benjamin-Lee/deep-rules/issues/20), [#19](https://github.com/Benjamin-Lee/deep-rules/issues/19))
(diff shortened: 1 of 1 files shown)


evancofer:
Though certainly relevant to 7, the placement in 10 is no mistake. The reason that this belongs in 10 is because it may be easier to recover private training data from overfit models with more parameters / greater representation capacity.

rules.md
```diff
@@ -4,6 +4,7 @@
   - Garbage In/Garbage Out ([#26](https://github.com/Benjamin-Lee/deep-rules/issues/26))
   - Normalize your data ([#7](https://github.com/Benjamin-Lee/deep-rules/issues/7))
   - Make sure data is not biased/skewed ([#43](https://github.com/Benjamin-Lee/deep-rules/issues/43))
+  - How to deal with class imbalance
```
rasbt: I am assuming this refers to bias/skew in terms of "unfairness" so I think adding a separate point on class imbalance makes sense (this issue often comes up in e.g., molecular activity analysis where there are lots of inactive molecules assayed and only a handful of actives)

rasbt:
Hm, interesting point. A counter-argument would be that due to the model complexity, everything becomes more "muddled" and it's not immediately obvious what the input/output relationships are if someone was to reconstruct some data based on having access to the model. 

For sure, I would say that traditional methods like KNN are much more prone to privacy leaks as they literally store the training dataset.

evancofer:
This is well outside my wheelhouse, but I am under the impression that relationships between the input and output don't need to be determined for a serious breach to occur. Just recovering a small amount of either one could be enough to de-anonymize a patient. This sort of risk must be taken seriously in biomedical applications.

rasbt:
> LGTM.
> 
> This is well outside my wheelhouse, but I am under the impression that relationships between the input and output don't need to be determined for a serious breach to occur. Just recovering a small amount of either one could be enough to de-anonymize a patient. This sort of risk must be taken seriously in biomedical applications.

I agree. And also, it's always better to err on the "being overly careful side" :). I added the "capacity" point under both 7) and 10) so that we can discuss it there in separate contexts.

PR #83, opened 2018-12-09
Update 02.intro.md

Benjamin-Lee:
To get the writing rolling, I've drafted a short introduction to the paper.

PR #91, opened 2018-12-11
Initial draft of Rule 1: Deep learning is still machine learning

evancofer:
This is an initial draft of Rule 1. I have mostly focused on a high level discussion of biased data and validating models (since that is what's mentioned in `rules.md`). I am hoping that this will initiate discussion of this rule, since we will inevitably need to heavily pare down general ML advice to find the subset that is most relevant or critical for a broad range of DL applications in computational biology.

### DIFF
--- a/content/03.ml-concepts.md
+++ b/content/03.ml-concepts.md
@@ -3,12 +3,12 @@ Deep learning is a distinct subfield of machine learning, but it is still a subf
 Deep learning has proven to be an extremely powerful paradigm capable of outperforming “traditional” machine learning approaches, but it is not immune to the many limitations inherent to machine learning.
 Many best practices for machine learning apply to deep learning as well.
 For instance, deep supervised learning models should be trained, tuned, and tested on non-overlapping datasets.
-Those developing deep learning models should first investigate their data.
+Those developing deep learning models should select data that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions.
+Furthermore, investigators should begin by thoroughly inspecting their data.
 When coupled with imprudence, data that is biased, skewed, or of low quality will produce models of dubious performance and limited generalizability.
 Biases in testing data can also unduly influence measures of model performance.
 For example, many conventional metrics for classification (e.g. area under the receiver operating characteristic curve or AUROC) have limited utility in cases of extreme class imbalance.
-As such, model performance should be evaluated with a carefully-picked panel of relevant metrics that make minimal assumptions about the composition of the testing data.
+As such, model performance should be evaluated with a carefully-picked panel of relevant metrics that make minimal assumptions about the composition of the testing data [@doi:10.1021/acs.molpharmaceut.7b00578].
 Extreme cases warrant testing the robustness of the model and metrics on simulated data for which the ground truth is known.
 Said simulations can be used to verify the correctness of the model’s implementation as well.
 Like all computational methods, deep learning should be leveraged in a systematic manner that is reproducible and rigorously tested.
-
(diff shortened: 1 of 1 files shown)


content/03.ml-concepts.md
```diff
@@ -1,2 +1,14 @@
 ## Rule 1: Concepts that apply to machine learning also apply to deep learning
+Deep learning is a distinct subfield of machine learning, but it is still a subfield.
```
rasbt: If we don't mention it in a prior introduction, I suggest that we also briefly mention what deep learning "is." Maybe sth like

"While deep learning is a distinct subfield of machine learning, that is particularly focused on deep artificial neural networks, it is still a subfield."

content/03.ml-concepts.md
```diff
@@ -1,2 +1,14 @@
...
+Deep learning has proven to be an extremely powerful paradigm capable of outperforming “traditional” machine learning approaches, but it is not immune to the many limitations inherent to machine learning.
+Many best practices for machine learning apply to deep learning as well.
+For instance, deep supervised learning models should be trained, tuned, and tested on non-overlapping datasets.
+Those developing deep learning models should first investigate their data.
+When coupled with imprudence, data that is biased, skewed, or of low quality will produce models of dubious performance and limited generalizability.
```
rasbt: I would also add "non-salient information" or sth like this. Because data can be high-quality, non-biased, non-skewed but if the information is not useful for the problem at hand, it may still result in poor models

content/03.ml-concepts.md
```diff
@@ -1,2 +1,14 @@
...
+Biases in testing data can also unduly influence measures of model performance.
+For example, many conventional metrics for classification (e.g. area under the receiver operating characteristic curve or AUROC) have limited utility in cases of extreme class imbalance.
```
rasbt: Maybe a good paper to cite here: 

Korotcov A, Tkachenko V, Russo DP, Ekins S: Comparison of Deep Learning With Multiple Machine Learning Methods and Metrics Using Diverse Drug Discovery Data Sets. Molecular Pharmaceutics 2017, 14:4462–4475.

content/03.ml-concepts.md
```diff
@@ -1,2 +1,14 @@
...
+As such, model performance should be evaluated with a carefully-picked panel of relevant metrics that make minimal assumptions about the composition of the testing data.
```
rasbt: Maybe a good paper to cite here:

Korotcov A, Tkachenko V, Russo DP, Ekins S: Comparison of Deep Learning With Multiple Machine Learning Methods and Metrics Using Diverse Drug Discovery Data Sets. Molecular Pharmaceutics 2017, 14:4462–4475.

content/03.ml-concepts.md
Benjamin-Lee: The current first line of the intro is "Deep learning (DL), a subfield of machine learning (ML) that focuses on deep artificial neural networks, is exploding in popularity and is increasingly being used for biological data analysis"

rasbt: Oh, I forgot that I already suggested it for the intro in https://github.com/Benjamin-Lee/deep-rules/pull/83. I knew it sounded familiar :D. Never mind this suggestion here then; would be redundant

SiminaB:
Great start! Should we mention the general uses of deep learning (which are the same as those of general machine learning) here or in the intro? For example, the fact that methods exist both for supervised and unsupervised learning. (My own pet peeve is that often people say that they want to predict a bunch of stuff but then it turns out they either can't predict based on their data or prediction is not actually the problem of interest in their case. I might try to include that somewhere, maybe in https://github.com/Benjamin-Lee/deep-rules/blob/master/content/06.know-your-problem.md)

evancofer:
@SiminaB @rasbt  I think that a definition of deep learning and machine learning needs to appear very early in the paper, and should probably go in the intro.

content/03.ml-concepts.md
evancofer: @rasbt I believe that this has been addressed with the most recent commit.

content/03.ml-concepts.md
evancofer: I agree. Added in the most recent commit.

SiminaB:
Sorry I missed the review request here (wasn't used to getting these)! Glad you got the approvals!

PR #89, opened 2018-12-11
Initial draft of Rule 10: Don't share models trained on sensitive data

AlexanderTitus:
Initial draft of [Rule 10](https://github.com/Benjamin-Lee/deep-rules/issues/4)

PR #92, opened 2018-12-11
Add Alexander Titus

Benjamin-Lee:


agitter:
@AlexanderTitus added himself in #89.  Should we close this and merge the update there along with the rule text?

Benjamin-Lee:
My mistake, yes

PR #103, opened 2018-12-19
Switch to Quick Tips

Benjamin-Lee:

**Any more details?**
For now, I'm not changing the repository name from `deep-rules` to `deep-tips` because I'm worried that might break either Travis or the GitHub Pages site. More details [here](https://blog.github.com/2013-05-16-repository-redirects-are-here/). @agitter, @dhimmel, can/should I change it?


README.md
```diff
@@ -11,23 +11,22 @@
...
+into ten quick tips, we hope to increase the number of biological researchers
 using DL (by making it more inviting) and the quality of the research (by
 helping them avoid common mistakes).
 
 Join us!
 
 ## Where we are now
 
-Currently, we're in the process of drafting the paper with each rule being tracked in its own [project](https://github.com/Benjamin-Lee/deep-rules/projects). To contribute to the paper, take a look at [`CONTRIBUTING.md`](CONTRIBUTING.md) for information on how we're organizing the writing as well as [`USAGE.md`](USAGE.md) for instructions on how to write the actual text.
+Currently, we're in the process of drafting the paper with each tip being tracked in its own [project](https://github.com/Benjamin-Lee/deep-rules/projects). To contribute to the paper, take a look at [`CONTRIBUTING.md`](CONTRIBUTING.md) for information on how we're organizing the writing as well as [`USAGE.md`](USAGE.md) for instructions on how to write the actual text.
```
agitter: This would also be a good time to note in the readme that the pre-submission inquiry was approved with a link to the issue.

content/02.intro.md
```diff
@@ -6,5 +6,5 @@ While much has been written about the impressive results of DL in biology, there
 
 To fix this problem, we solicited input from a diverse community of researchers, who wrote this manuscript collaboratively using the GitHub version control platform [@url:https://github.com/Benjamin-Lee/deep-review] and Manubot [@url:https://greenelab.github.io/meta-review/].
 In the course of our discussions, several themes became clear: the importance of understanding and applying ML fundamentals as a baseline for utilizing DL, the necessity for extensive model comparisons and careful evaluation, and the need for critical thought in interpreting results generated by means of DL, among others.
-Ultmately, the rules we established range from high-level guidance to the implementation of best practices, and it is our hope that they will provide actionable, DL-specific advice for both new and experienced DL practitioners alike who would like to employ DL in biological research.
+Ultmately, the tip we established range from high-level guidance to the implementation of best practices, and it is our hope that they will provide actionable, DL-specific advice for both new and experienced DL practitioners alike who would like to employ DL in biological research.
```
agitter:
```suggestion
Ultmately, the tips we established range from high-level guidance to the implementation of best practices, and it is our hope that they will provide actionable, DL-specific advice for both new and experienced DL practitioners alike who would like to employ DL in biological research.
```

dhimmel:
> can/should I change it?

Travis may need to be reconfigured.

If you rename, but then create a repository at `Benjamin-Lee/deep-rules`, redirects will not work obviously.

Finally, the old permalinks will break:

> There is one caveat with the new redirect support worth noting: GitHub Pages sites are not automatically redirected when their repositories are renamed at this time.

That is probably the worst part, so if you are going to rename, sooner is better than later.

cgreene:
I would advocate for not renaming, since the existing links would break. The name is close enough to the goal, that with an updated readme I wouldn't worry about it.

pstew:
Agreed, I don't think we should rename.

Benjamin-Lee:
Ok no renaming it is. I’ll add a quick description to the README when I get back to my laptop.

Benjamin-Lee:
@agitter Take a look now!

PR #104, opened 2018-12-19
Initial draft of text for Rule #6 (Hyperparameter optimization)

cgreene:

**Any more details?**

I wanted to get some text down on this rule so that the blank page effect becomes less intimidating. I'm expecting this will need some substantial revision and re-working (especially with regards to rule 2 once a draft of that lands).

### DIFF
--- a/content/citation-tags.tsv
+++ b/content/citation-tags.tsv
@@ -1,3 +1,3 @@
 tag	citation
 deep_review	doi:10.1101/142760
-hornik	10.1016/0893-6080(91)90009-T
+hornik-approximation	10.1016/0893-6080(91)90009-T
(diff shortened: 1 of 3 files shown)


content/citation-tags.tsv
```diff
@@ -1,2 +1,3 @@
 tag	citation
 deep_review	doi:10.1101/142760
+hornik	10.1016/0893-6080(91)90009-T
```
agitter:
```suggestion
hornik	doi:10.1016/0893-6080(91)90009-T
```

I also recommend making the tag slightly more description with a date or keyword, like `hornik1991` or `hornik-approximation`.

Issue #82, opened 2018-12-09
Have access to adequate infrastructure


rasbt:
**Have you checked the [list of proposed rules](https://github.com/Benjamin-Lee/deep-rules/issues?q=is%3Aissue+is%3Aopen+label%3Arule) to see if the rule has already been proposed?**

- [x] Yes

**Did you add yourself as a [contributor](https://github.com/Benjamin-Lee/deep-rules/blob/master/contributors.md) by making a pull request if this is your first contribution?**

- [x] Yes, I added myself or am already a contributor

**Feel free to elaborate, rant, and/or ramble.**

As far as I know, we don't have anything about infrastructure demands, yet. I.e., using DL in ones research will likely require access to multiple GPUs etc and is usually of higher computational demand compared to traditional models. So, researchers who relied on laptops and off-the-shelf desktop workstations for computational analyses should made aware that DL usually requires a bit more :).

This could potentially be mentioned in several places, e.g., 

2. Use traditional methods to establish performance baselines ([#41](https://github.com/Benjamin-Lee/deep-rules/issues/41), [#11](https://github.com/Benjamin-Lee/deep-rules/issues/11), [#10](https://github.com/Benjamin-Lee/deep-rules/issues/10))
- 3. Understand the complexities of training deep neural networks
- 6. Tune your hyperparameters extensively and systematically ([#42](https://github.com/Benjamin-Lee/deep-rules/issues/42), [#11](https://github.com/Benjamin-Lee/deep-rules/issues/11))

**Any citations for the rule?** (peer-reviewed literature preferred but not required)
- [DOI](doi.org/DOI_goes_here)

evancofer:
It may be worth pointing out programs like [NVIDIA's GPU grant program](https://developer.nvidia.com/academic_gpu_seeding). These can be really useful for a small project or for new labs interested in DL. I believe some of the academic cloud credit programs can also be used for GPU or TPU compute time.

rasbt:
Good point.  NVIDIA's GPU grant program is definitely nice and worth mentioning for smaller projects. Regarding the cloud credit, that's another thing to consider. I think AWS is mainly for teaching, but I am not sure. There are probably dozens of alternative programs that exist. We may want to mention this in general, because it could be immensely useful to researchers who think about employing DL but don't want to make a huge investment where it is unsure if DL really helps in their projects.

fmaguire:
Seems best fit for the training complexities

[...]

Issue #63, opened 2018-11-29
Future-proof: Plan how you will support and share your model


smsaladi:
**Have you checked the [list of proposed rules](https://github.com/Benjamin-Lee/deep-rules/issues?q=is%3Aissue+is%3Aopen+label%3Arule) to see if the rule has already been proposed?**

- [X] Yes

**Did you add yourself as a [contributor](https://github.com/Benjamin-Lee/deep-rules/blob/master/contributors.md) by making a pull request if this is your first contribution?**

- [ ] Yes, I added myself or am already a contributor

Much of the discussion this far has concerned with creating/crafting a model, but as far as I can tell, there hasn't been much discussion on what happens afterwards (perhaps, let's say, after you publish the model). Since lots of deep learning is done to solve problems, making the model available as a tool seems just as important as crafting it.

This conversation is not unique to deep learning, but I think it's especially relevant because the hard work being done to make the application layer (Tensorflow, pytorch) runnable across machines/architectures and to support it.

In the past, there's been a question of "can you get everything running on your machine", say to process the input data in the format necessary for inference (in my experience, this is not trivial) and then actually run the inference code. However, with deep learning, often times the input is the raw data itself (perhaps with a tad bit of processing). In my experience with sequence bioinformatics, people are building models that use the DNA or protein sequence itself, without any sort of featurization calculations/pre-processing.

With a deep learning model, if it's built on a "standard" platform, sharing the model and running it oneself is significantly simplified.

When building a deep learning model (or perhaps shortly after), my suggestion is that the authors think about how they will share it with the user community. 
* Will they provide the model, weights, and a code example of how to use it for inference (least convenient)? 
* Will they run a web application that takes in data, runs the inference, and then returns the output to the user in a easy-to-understand way (and where will they run this? heroku? locally?)? 
* Will they have a database of precomputed inputs and provide this to the user (simply as data? or though a web api?)?
* Will they create a static website that can take input data and run inference in the user's browser itself  and render the output there itself? There's lots of work being done (e.g. [TensorFlowJS](https://js.tensorflow.org/), [TensorRT](https://developer.nvidia.com/tensorrt)) on the inference stage/lifetime of a model. (possibly the most convenient in the long term) 

And, how will they maintain the application going forward? What if they (or someone else) puts out a model that's performs better at the same benchmarks? Will they turn their service off if the second model is made available? Will they host this second model?

PS. This is an awesome effort!

rasbt:
[...]

Issue #52, opened 2018-11-21
Perform sanity checks and follow good coding practices


evancofer:
**Have you checked the [list of proposed rules](https://github.com/Benjamin-Lee/deep-rules/issues?q=is%3Aissue+is%3Aopen+label%3Arule) to see if the rule has already been proposed?**

- [x] Yes

**Did you add yourself as a [contributor](https://github.com/Benjamin-Lee/deep-rules/blob/master/contributors.md) by making a pull request if this is your first contribution?**

- [x] Yes, I added myself or am already a contributor

**Feel free to elaborate, rant, and/or ramble.**
When coding DL models, it is important to maintain good software engineering practices. All code should be documented and include rigorous tests. Sanity checks are also useful. For instance, something is probably wrong (e.g. bug in code, ill posed problem, bad hyperparameters) if model training loss does not decrease (i.e. not overfitting) when considering a very small subset of the training data.

**Any citations for the rule?** (peer-reviewed literature preferred but not required)
- https://arxiv.org/pdf/1206.5533v2.pdf
- http://cs231n.github.io/neural-networks-3/#sanitycheck

evancofer:
This slightly subsumes #49

rasbt:
I think it would be good to have your suggestion as s a separate rule, but it is also somewhat connected #42, the fact that we need to usually have a larger/more extensive model selection part when using deep learning as opposed to "traditional" machine learning. In addition, we need to "more babysit" the different model fitting procedures and evaluating the internal procedure ("does it converge?") vs just the external metrics ("what is the prediction accuracy?")

evancofer:
Indeed. In many cases however, sanity checks need to occur before hyperparameter optimization occurs.

Benjamin-Lee:
>  All code should be documented and include rigorous tests.

With respect to this, I just published [a paper](doi.org/10.1371/journal.pcbi.1006561) on the exact topic that might be worth citing. I'm happy to expand further on documentation best practices for DL.

agitter:
I also recommend [*Top considerations for creating bioinformatics software documentation*](https://doi.org/10.1093/bib/bbw134) for software documentation

pstew:
@Benjamin-Lee Congrats! I think this and the paper linked by @agitter would be great to cite here.

fmaguire:
Sort of mentioned in tip 3, need to add this reference to it.

Issue #44, opened 2018-11-15
Expect to revise your models as new (or improved) datasets and methods become available


biomadeira:
**Have you checked the [list of proposed rules](https://github.com/Benjamin-Lee/deep-rules/issues?q=is%3Aissue+is%3Aopen+label%3Arule) to see if the rule has already been proposed?**

- [x] Yes

**Feel free to elaborate, rant, and/or ramble.**

Despite all our best efforts in trying to use the the cleanest of the datasets, balanced but heterogenous data, optimal hyperparameter settings, etc., we should expect to revise our models as new data and methods become available. We have to assume that our current set of observations is incomplete and our understanding is limited. Larger and improved datasets, metadata and other annotations will inevitably become available as new experimental techniques are introduced. This is particularly true across prolific fields such as regulatory genomics, transcriptomics, systems biology and biological image analysis (Angermueller et al., 2016, Ching et al., 2018). It will mean that models need to be re-trained and re-tested. Deep learning is still a young technology and in addition to new data, we should expect new developments and improvements to the methods and libraries we currently use.  

**Any citations for the rule?** (peer-reviewed literature preferred but not required)
- Angermueller et al., 2016. [doi.org/10.15252/msb.20156651](https://doi.org/10.15252/msb.20156651)
- Ching et al., 2018. [doi.org/10.1098/rsif.2017.0387](https://doi.org/10.1098/rsif.2017.0387)

rasbt:
Same is maybe true to some extend for architecture improvements. Just a trivial example, but as BatchNorm became kind of established in the field, it could be useful to try adding it to your algorithm and retrain to your data. Or say you used a ResNet-34 architecture and then later came a ResNet-50 with the bottleneck block that might further improve things. Of course, I wouldn't recommend trying every possible new thing out there, but there are several obvious advancements/"tricks" that usually help.

fmaguire:
Mentioned in https://github.com/Benjamin-Lee/deep-rules/blob/master/content/05.dl-complexities.md

PR #124, opened 2019-01-26
Tip 3 draft

fmaguire:

**Any more details?**

Draft of tip 3, with points from #42 #82 #44 #52 #63 

Could be expanded to cover more of #15 

Wasn't sure how much of #22 to add without being redundant to Tip 4.


content/05.dl-complexities.md
```diff
@@ -1 +1,25 @@
 ## Tip 3: Understand the complexities of training deep neural networks {#complexities}
+
+Correctly training deep neural networks is a non-trivial process.
+There are many different options and potential pitfalls at every stage.
+To get good results you have to expect to train many networks with
+different tweaks and modifications.
+Despite improved framework ease-of-use and on-demand cloud computing resources
+this means DL can be very demanding, requiring significant
+infrastructure and patitence to complete.
```
rasbt: "patitence" => "patience"

content/05.dl-complexities.md
```diff
@@ -1 +1,25 @@
...
+The experimentation inherent to DL is often noisy (requiring repetition) and
+represent a significant organisational challenge.
```
rasbt: "organisational" => "organizational"

(sorry, don't want to be nit-picky, but we used AE elsewhere)

content/05.dl-complexities.md
```diff
@@ -1 +1,25 @@
...
+All code, parameters, and results must be carefully corralled using good
+coding practices (e.g. version control, continuous integration etc.) in order to
```
rasbt: e.g. and etc. shouldn't be used in the same statement. I suggest

- (e.g., version control and continuous integration)

or 

- (for example, version control and continuous integration)

content/05.dl-complexities.md
```diff
@@ -1 +1,25 @@
...
+different tweaks and modifications.
+Despite improved framework ease-of-use and on-demand cloud computing resources
+this means DL can be very demanding, requiring significant
+infrastructure and patitence to complete.
+The experimentation inherent to DL is often noisy (requiring repetition) and
+represent a significant organisational challenge.
+All code, parameters, and results must be carefully corralled using good
+coding practices (e.g. version control, continuous integration etc.) in order to 
+be effective and interpretable.
+This organisation is also key to being able to efficiently share your work
```
rasbt: "organisation" => "organization"

content/05.dl-complexities.md
fmaguire: Fixed

content/05.dl-complexities.md
fmaguire: Not at all, totally automatic Britishism on my behalf.

content/05.dl-complexities.md
```diff
@@ -1 +1,25 @@
...
+and to update your model as new data becomes available. 
+
+
+Similar to [Tip 4](#baselines), try to start with a relatively smaller network
```
rasbt: Not sure if we were 100% consistent about that, but I think we referred to the reader in 3rd person before ("the researcher", "the experimenter," etc.)

content/05.dl-complexities.md
fmaguire: Fixed.

content/05.dl-complexities.md
fmaguire: Might be easiest to fix that flow/match tone when we have a draft for all sections?

content/05.dl-complexities.md
```diff
@@ -1 +1,25 @@
...
+and increase the size and complexity as needed to prevent wasting time and 
+resources. 
+Beware of the seemingly trivial choices that are being made implicitly by 
+default settings in your framework of choice e.g. choice of optimization algorithm
+(adaptive methods are often faster but can lead to worse results [@arxiv:1705.08292]).
```
rasbt: > "adaptive methods are often faster but can lead to worse results"

I'd suggest

"adaptive methods often lead to faster convergence during training but may lead to worse generalization performance on independent datasets" 

@arxiv:1705.08292: This was a 2017 NIPS paper, maybe it would be worth citing the peer-reviewed version:

Wilson, A. C., Roelofs, R., Stern, M., Srebro, N., & Recht, B. (2017). The marginal value of adaptive gradient methods in machine learning. In Advances in Neural Information Processing Systems (pp. 4148-4158).

A related one is https://arxiv.org/abs/1712.07628 by Keskar & Socher, but not sure if this has been published somewhere

content/05.dl-complexities.md
rasbt: I agree. Right now, these things are a bit difficult to address without having the context of all chapters!

content/05.dl-complexities.md
fmaguire: Think I managed to add the json format properly... guess I'll find out in a second once travis finishes...

fmaguire: Updated to Wilson citation from the NeurIPS proceedings

content/05.dl-complexities.md
rasbt: Sounds good. Like you suggested before, it may be better to wait until we have all the tips outlined in a first draft before we smooth it out to make it consistent :)

content/05.dl-complexities.md
agitter: For NeurIPS proceedings, URL citations are an option so that you don't have to manually provide the JSON: `url:https://papers.nips.cc/paper/7003-the-marginal-value-of-adaptive-gradient-methods-in-machine-learning`

It doesn't work well with every conference proceedings site, but this paper looks accurate.

content/05.dl-complexities.md
```diff
@@ -1 +1,25 @@
 ## Tip 3: Understand the complexities of training deep neural networks {#complexities}
+
+Correctly training deep neural networks is a non-trivial process.
+There are many different options and potential pitfalls at every stage.
+To get good results you have to expect to train many networks with
```
agitter: Can you please put a single sentence on each line instead of using hard line wrapping?  That makes it easier to track and comment on each sentence.

fmaguire: Sure, that makes sense. Done.

content/manual-references.json
```diff
@@ -18,5 +18,74 @@
       "family": "Himmelstein"
     }
   ]
-}
-]
\ No newline at end of file
+},
+{
```
agitter: If the URL citation is working now, you can please revert this file to remove the manual reference.  Thanks.

fmaguire: Sure thing, done now.

content/05.dl-complexities.md
```diff
@@ -1 +1,14 @@
 ## Tip 3: Understand the complexities of training deep neural networks {#complexities}
+
+Correctly training deep neural networks is a non-trivial process.
+There are many different options and potential pitfalls at every stage.
+To get good results you have to expect to train many networks with different tweaks and modifications.
+Despite improved framework ease-of-use and on-demand cloud computing resources this means DL can be very demanding, requiring significant infrastructure and patience to complete.
+The experimentation inherent to DL is often noisy (requiring repetition) and represents a significant organizational challenge.
+All code, parameters, and results must be carefully corralled using good coding practices (for example, version control, continuous integration etc.) in order to be effective and interpretable.
```
agitter: I'm curious about examples of continuous integration with DL.  Is there anything we could refer to?  Given that model training and performance can depend on things like the hardware and versions of required packages, I am interested in best practices here and think our readers would be as well.

content/05.dl-complexities.md
```diff
@@ -1 +1,14 @@
...
+This organization is also key to being able to efficiently share your work and to update your model as new data becomes available. 
+
```
agitter: Extra linebreak here

content/05.dl-complexities.md
rasbt: We definitely to mention non-deterministic algorithm choices in CUDA & cuDNN. I.e., by default in e.g., TensorFlow and PyTorch, running code on the GPU (as opposed to CPU) is non-deterministic usually (esp. convolutions). This is because cuDNN has many different approximations of convolutions implemented (i.e., all the winnograd stuff and derivates thereof approximating convolutions via fast fourier transforms). So, by default, if you run GPU code, the "best" algorithm is chosen at runtime and may vary from run to run (and can also change during runtime I think). Hence, it's best practice for reproducibility to set `torch.backends.cudnn.deterministic = True` in PyTorch, for example (forgot how to do that in TensorFlow). In practice, it's also often faster as the "guessing" by cuDNN is not that good (heard that from someone at NVIDIA as well when I reported it). 

Note that setting random seeds will only affect things like deterministic shuffling and weight initialization, the algorithm determinism has to be set on top of that. I think this is important to mention. We could reference the cuDNN technical manuals for this: https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#reproducibility

content/05.dl-complexities.md
```diff
@@ -1 +1,14 @@
 ## Tip 3: Understand the complexities of training deep neural networks {#complexities}
+
+Correctly training deep neural networks is a non-trivial process.
+There are many different options and potential pitfalls at every stage.
+To get good results you have to expect to train many networks with different tweaks and modifications.
```
rasbt: > "with different tweaks and modifications." 

 "tweaks and modifications" sounds good, but we should also say "settings" because changing things like the learning rate is not necessarily a tweak or modification (I think of tweaks and modifications more of larger changes, like adding dropout, batchnorm, varying the number of layers, and so forth)

content/05.dl-complexities.md
```diff
@@ -1 +1,14 @@
...
+Despite improved framework ease-of-use and on-demand cloud computing resources this means DL can be very demanding, requiring significant infrastructure and patience to complete.
```
rasbt: I think we need to add a comparison here, because that's of course very relative. Just attended a seminar last week where someone ran simulations on 4096 CPUs that took months to complete ... the throughput was 1 petaflop, which could have been easily achieved via a handful of GPUs -- what I am trying to say is that the infrastructure is relatively demanding compared to maybe traditional machine learning on small datasets. Also, traditional MCMC or all the molecular dynamics research is much more expensive than DL (e.g., I would say DL is currently probably the cheapest method for protein structure refinement compared to traditional methods :P)

rasbt:
> @rasbt did you have any other feedback?

I just added some additional comments we may want to address, i.e., 

- including sth like "settings" when talking about "requiring modifications"
- adding reference for comparison when saying that DL is demanding
- the reproducibility aspect with GPU computing

we could add this later, but since everything could be addressed on 0.5 -2 sentences, I think it's organizationally easier to do that before we merge (if you agree with these points, of course). Otherwise and overall, it looks pretty good to me :)

agitter:
Yes, I agree with your points above.  Because they are fairly specific suggestions, I also think we can make these changes before merging.

content/05.dl-complexities.md
fmaguire: I couldn't really find a good general comparison but I did find a paper that was looking at scaling and computational performance.  We don't need to say necessarily it is WORSE than other methods to have a strong case that it can still require significant infrastructure etc.

content/05.dl-complexities.md
fmaguire: Couldn't really find a good CI example specifically for DL usage instead of writing new modules/frameworks.   

Good call on the non-deterministic/deterministic backend algorithms part, I'd forgotten totally about that.

PR #130, opened 2019-02-05
Update contributors.md

AlexanderTitus:

- [X] Yes, I added myself or am already a contributor

**Any more details?**

This is to update the contributors page to acknowledge accountability for the future iterations and accuracy of the work.

### DIFF
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -17,3 +17,11 @@ author_info:
       - School of Engineering and Applied Sciences, Harvard University
       - Department of Genetics, Harvard Medical School
       - Lab41, In-Q-Tel
+  -
+    github: AlexanderTitus
+    name: Alexander J. Titus
+    initials: AJT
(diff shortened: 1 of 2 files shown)


content/metadata.yaml
```diff
@@ -17,3 +17,11 @@ author_info:
       - School of Engineering and Applied Sciences, Harvard University
       - Department of Genetics, Harvard Medical School
       - Lab41, In-Q-Tel
+  -
+    github: AlexanderTitus
```
agitter: @Benjamin-Lee do you want contributors/authors to start adding themselves to `metadata.yaml` now or after the ordering has been determined?  The order in this file will be the order in the generated HTML and PDF.

agitter: To follow up, I'm not questioning whether @AlexanderTitus should be an author.  His contributions to the project are quite clear.  The question was only about the timing of listing authors.

Issue #42, opened 2018-11-14
Be patient and pay attention to seemingly trivial hyperparameter settings


rasbt:
**Have you checked the [list of proposed rules](https://github.com/Benjamin-Lee/deep-rules/issues?q=is%3Aissue+is%3Aopen+label%3Arule) to see if the rule has already been proposed?**

- [x] Yes

In general, getting DL models to work on even simple, structured datasets requires much extensive hyperparameter tuning (Koutsoukas et al. 2017) [Could also add a rule saying that a 2-layer multi-layer perceptron is not deep learning ;)] compared to "traditional" machine learning. Hence, it is important to be patient and try many different hyperparameter settings and combinations.

For instance, the previously "best" learning rate might become useless if we add another layer or change the activation function. Hence, extensive tuning and a near-exhaustive search is recommended.

Furthermore, it should be stressed that even the same architecture and hyperparameter configuration should be tested multiple times with different random seeds since random weight initialization may make a difference between convergence and non-convergence. For model selection and evaluation, it is recommended is to evaluate models based on their average performance based on at least a "top 3 out of 5 models for a given hyperparameter setting" comparison.

**Any citations for the rule?** (peer-reviewed literature preferred but not required)
- Koutsoukas et al. 2017 [DOI](https://doi.org/10.1186/s13321-017-0226-y)

agitter:
A corollary of this rule is that DL is compute-intensive.  Be prepared to train many models when starting a project.

rasbt:
That's a nice way to put it! Basically, we want to highlight that when using DL (as opposed to e.g. a deterministic KNN classifier, decision tree, SVM, etc.), the hyperparameter tuning is exponentially/way more expensive, because

- there are usually more hyperparameters to search over (esp. if you take architecture modifications into account)
- each hyperparameter setting should be tried multiple times with different random seeds, because, the models may converge to different local minima or don't converge at all, so some models will be more useful than others, and some will not be useful at all even if all the settings are the same.

evancofer:
We might want to add something about effective but simple hyperparameters/architecture modifications to experiment with (e.g. dropout, batch normalization). It may also be worth mentioning that adaptive learning methods (e.g. Adam) can help save time when determining a viable model architecture (i.e. before even optimizing hyperparameters).

rasbt:
Sure :)  Maybe also adding skip connections to the list

fmaguire:
[...]

Issue #39, opened 2018-11-14
Exploit knowledge of similar data, use transfer learning!


Isy89:
**Have you checked the [list of proposed rules](https://github.com/Benjamin-Lee/deep-rules/issues?q=is%3Aissue+is%3Aopen+label%3Arule) to see if the rule has already been proposed?**

- [x ] Yes

**Feel free to elaborate, rant, and/or ramble.**

**Any citations for the rule?** (peer-reviewed literature preferred but not required)
- [DOI](doi.org/DOI_goes_here)

cgreene:
See #8

Issue #28, opened 2018-11-11
Understand, protect against, and benchmark possible overfitting


chevrm:
**Have you checked the [list of proposed rules](https://github.com/Benjamin-Lee/deep-rules/issues?q=is%3Aissue+is%3Aopen+label%3Arule) to see if the rule has already been proposed?**

- [ X] Yes

**Feel free to elaborate, rant, and/or ramble.**

**Any citations for the rule?** (peer-reviewed literature preferred but not required)
- [DOI](doi.org/DOI_goes_here)

evancofer:
It is probably worth mentioning dropout ([discussed here](https://www.cs.toronto.edu/~hinton/absps/JMLRdropout.pdf)) and weight decay, both of which are extremely good for shrinking the gap between training and validation performance.

rasbt:
Just throwing in some references ... a general one regarding easily picking up spurious correlations (or rather systematic noise) is

- Jo, J., & Bengio, Y. (2017). Measuring the tendency of CNNs to Learn Surface Statistical Regularities. arXiv preprint arXiv:1711.11561.

(There were some other ones similar to this that I can't recall top of my head.)

Regarding correlation of training set size and performance in a bio application:

- Mayr A, Klambauer G, Unterthiner T, Steijaert M, Wegner JK, Ceulemans H, Clevert D-A, Hochreiter S: Large-scale comparison of machine learning methods for drug target prediction on ChEMBL. Chemical Science 2018, 9:5441–5451.

although they found that this was equally true for traditional ML. So maybe we need to add a general DL paper to highlight to be careful when deciding upon traditional vs DL when training sets are small.

A bio-related one

- Koutsoukas A, Monaghan KJ, Li X, Huan J: Deep-learning: investigating deep neural networks hyper-parameters and comparison of performance to shallow methods for modeling bioactivity data. Journal of Cheminformatics 2017, 9.

where they found that naive Bayes was better for noisy datasets

agitter:
There is a lot to like in that Mayr et al. 2018 paper.  They were very thorough in exploring different models, hyperparameters, and chemical featurizations.  However, they only consider ROC as an evaluation metric, and most of the ChEMBL targets are highly class-imbalanced.

Issue #8, opened 2018-11-01
If you don't have enough data, consider whether there is other data that can help pre-train a model.


cgreene:
Often we lack a large amount of labeled data, but sometimes there are abundant unlabeled data. We could touch on pre-training approaches (autoencoder pre-training, etc) as well as transfer + refinement (though you have that in #6 so transfer could just be added there). Perhaps this ends up subsumed into a broader point as people contribute their tips.

julianstanley:
I think it would be great to combine this issue with #6 into a more general heading, maybe along the lines of "Don't reinvent the wheel". That can include things like:

1. Pretraining (as is mentioned in this comment), we can site Razavian ([arxiv](https://arxiv.org/abs/1403.6382)) to support that pretraining is useful, but cite something like Yosinski ([nips](http://papers.nips.cc/paper/5347-how-transferable-are-features-in-deep-n%E2%80%A6)) to remind that pretraining is definitely context-dependent.

2. Available tools (Tensorflow and other open source libraries are often expanding. Check to make sure that someone hasn't already built a tool that solves your problem)

3. Finding similar data. There is a lot of already-available biological data from places like GEO. Training (or perhaps more importantly, testing) a model on already-existing data may help increase the ability of the model to generalize to other contexts.

evancofer:
To many, one of the attractive features of DL is that it offers a high degree of flexibility (with respect to input feature representation) while remaining simple to implement and train. As such, DL may sometimes be a useful approach despite limited training data. In those cases, it is probably a good idea to use regularization methods (e.g. dropout, weight decay, input noise, etc).

PR #134, opened 2019-02-09
Tip 5 (architecture) initial draft

Benjamin-Lee:

**Any more details?**

To get something on the page. I changed to title to "Choose an appropriate data representation and neural network architecture" because I think data representation logically comes before architecture design.


SiminaB:
This tip is really interesting to me! I wonder if we could expand a bit on some parts of Table 1 of the review paper led by @cgreene  and @agitter https://royalsocietypublishing.org/doi/pdf/10.1098/rsif.2017.0387 For instance:

If you have imaging or DNA sequence data, consider CNNs, because they emphasize local structures/adjacency
If there is no natural ordering of inputs, like in gene expression data, consider MLP
If you have data with structure and longer-term dependencies, like natural language, consider LSTM
etc

It would be great to also provide some intuition on why, for example, CNNs emphasize local structures/adjacency (because of the convolution steps)

content/07.architecture-and-representation.md
```diff
@@ -1 +1,3 @@
-## Tip 5: Choose an appropriate neural network architecture and data representation {#architecture}
+## Tip 5: Choose an appropriate data representation and neural network architecture {#architecture}
+Unfortunately, choosing how to represent your data and design your architecture is closer to an art than a science.
+With that being said, there are some general principles that are useful to follow when experimenting.
```
rasbt: This is a nice start! Here a suggestion to extend this a bit:

Unfortunately, choosing how to represent your data and design your architecture is closer to an art than a science. While certain best practices have been established by the research community {cite see below}, architecture design choices remain largely problem-specific and vastly empirical efforts requiring extensive experimentation. Furthermore, as deep learning is a fast moving field, many recommendations are often short-lived and frequently replaced by newer insights supported by recent empirical results. This is further complicated by the fact that many recommendations are very problem- and dataset-specific and do generalize well across different tasks. With that being said, there are some general principles that are useful to follow when experimenting.

- I recommend citing the collection "Neural Networks: Tricks of the Trade" to which many established authors contributed their insights: https://doi.org/10.1007/978-3-642-35289-8

agitter: Well said!

Did you mean to write:
> very problem- and dataset-specific and do _**not**_ generalize

rasbt: Oh yes, definitely meant **not** generalize here :)

chevrm: Agreed. I like @rasbt 's extension.

alxndrkalinin:
I went over suggested issues and I think 2 of the most important ones that aren't described in the text are external data / transfer learning (#8, #39). Seems like https://github.com/Benjamin-Lee/deep-rules/issues/8#issuecomment-437986846 suggested to generalize and add it to this section. I'd love to write a couple of paragraphs if that's still of interest, possibly tying this to overfitting (#28) and careful tuning (#42). WDYT?

rasbt:
> I went over suggested issues and I think 2 of the most important ones that aren't described in the text are external data / transfer learning (#8, #39). Seems like [#8 (comment)](https://github.com/Benjamin-Lee/deep-rules/issues/8#issuecomment-437986846) suggested to generalize and add it to this section. I'd love to write a couple of paragraphs if that's still of interest, possibly tying this to overfitting (#28) and careful tuning (#42). WDYT?

Looking at the other section titles again, I think this one would be indeed the best fit for mentioning transfer learning. For overfitting, current Tip 7 (note the filename number is a bit different) would maybe be a better fit? I.e., https://github.com/Benjamin-Lee/deep-rules/blob/master/content/09.overfitting.md . The same is probably true for "tuning". I.e., would probably better fit into 

- https://github.com/Benjamin-Lee/deep-rules/blob/master/content/09.overfitting.md
- https://github.com/Benjamin-Lee/deep-rules/blob/master/content/05.dl-complexities.md

Benjamin-Lee:
@alxndrkalinin (and everyone else), feel free to commit any changes to this PR. I'll try to incorporate your ideas in upcoming commits.

alxndrkalinin:
I ended up creating a separate PR– #141, feel free to review / edit / make suggestions

Benjamin-Lee:
Thanks so much for the help @alxndrkalinin! I really appreciate it.

content/07.architecture-and-representation.md
```diff
@@ -1 +1,21 @@
-## Tip 5: Choose an appropriate neural network architecture and data representation {#architecture}
+## Tip 5: Choose an appropriate data representation and neural network architecture {#architecture}
+Unfortunately, choosing how to represent your data and design your architecture is closer to an art than a science.
+While certain best practices have been established by the research community [doi:10.1007/978-3-642-35289-8], architecture design choices remain largely problem-specific and vastly empirical efforts requiring extensive experimentation.
```
agitter:
```suggestion
While certain best practices have been established by the research community [@doi:10.1007/978-3-642-35289-8], architecture design choices remain largely problem-specific and are vastly empirical efforts requiring extensive experimentation.
```

content/07.architecture-and-representation.md
```diff
@@ -1 +1,21 @@
...
+Furthermore, as deep learning is a quickly evolving field, many recommendations are often short-lived and frequently replaced by newer insights supported by recent empirical results.
+This is further complicated by the fact that many recommendations are very problem- and dataset-specific and do not generalize well across different tasks.
```
agitter:
```suggestion
This is further complicated by the fact that many recommendations do not generalize well across different problems and datasets.
```

content/07.architecture-and-representation.md
```diff
@@ -1 +1,21 @@
...
+With that being said, there are some general principles that are useful to follow when experimenting.
+
+First and foremost, use your knowledge of the available data and your question (see [Tip 4](#know-your-problem)) to inform your data representation and architectural design choices.
+For example, if your data is an array of measurements with no natural ordering of inputs (such as gene expression data), multilayer perceptrons (MLPs), which are the most basic type of neural network, may be effective.
+Similarly, if your data is comprised of images, convolutional neural networks (CNNs) are a good choice because they emphasize local structures and adjacency within the data.
```
agitter:
```suggestion
Similarly, if your data is comprised of sequences or images, convolutional neural networks (CNNs) are a good choice because they emphasize local structures and adjacency within the data.
```

content/07.architecture-and-representation.md
```diff
@@ -1 +1,21 @@
...
+
+DL models can typically benefit from large amounts of labelled data to avoid overfitting (see [Tip 7](#overfitting)) and to achieve top performance on a task in hand.
+In the event that there is not enough data available to train your model, consider using transfer learning (TL).
```
agitter: In a short article like this, I would prefer to minimize the use of acronyms as much as possible.  I believe is makes it more accessible to non-specialists.  Could we replace `TL` with the full `transfer learning` in this tip?

alxndrkalinin: Agreed. Only reason I used it was because `Introduction` uses ML and DL acronyms throughout, which btw then aren't used as much in other chapters. May need to unify the use of those as well.

alxndrkalinin:
@agitter Thank you for the review and useful suggestions (never used this feature before–very cool!). I committed your edits, replaced TL with full name, and added my metadata.

agitter:
> never used this feature before–very cool!

The GitHub interface is changing rapidly.  Last week they also added [draft pull requests](https://github.blog/2019-02-14-introducing-draft-pull-requests/), which seem to replace the WIP plugin we've been using.

You're right that we'll need to be consistent with acronym avoidance elsewhere in the manuscript as well.

Benjamin-Lee:
Removing WIP tag. Let's get this merged into the main draft.

content/07.architecture-and-representation.md
```diff
@@ -1 +1,22 @@
...
+CNNs may also be a good choice for learning on sequences, as recent empirical evidence suggests they can outperform canonical sequence learning techniques such as recurrent neural networks (RNNs) and the closely related long short-term memory (LSTM) networks [arxiv:1803.01271]. 
+
+DL models can typically benefit from large amounts of labelled data to avoid overfitting (see [Tip 7](#overfitting)) and to achieve top performance on a task in hand.
+In the event that there is not enough data available to train your model, consider using transfer learning.
+In transfer learning, a model whose weights were generated by training on another dataset is used as the starting point for training [@tag:Yosinski2014].
+Transfer learning is most useful when pre-training and target datasets are of similar nature [@tag:Yosinski2014].
+For this reason, it is important to search for similar data that is already available and may potentially be used to increase the size of the training set or for pre-training and subsequent fine-tuning on the target data.
```
rasbt: In TL, we also don't always want/need to completely start from scratch and use a similar dataset to pre-train the model ourselves. Many pre-trained models are also off-the-shelf available for further training. Maybe we should add a comment about that? E.g., something like

"As an alternative to pre-training models on larger datasets for transfer learning yourself, you may also be able to obtain pre-trained models from public repositories."

Or is it better not to recommend that (as the publicly available models are mostly "too general")?

content/07.architecture-and-representation.md
```diff
@@ -1 +1,22 @@
...
+However, even when this assumption does not hold, transferring features still can improve performance of the model compared to just random feature initialization.
+For example Rojkomar et al. showed advantages of ImageNet-pretraining [@doi:10.1007/s11263-015-0816-y] for the model that is applied to greyscale medical image classification [@doi:10.1007/s10278-016-9914-9].
```
rasbt: AE again :) "greyscale" -> "grayscale"

content/07.architecture-and-representation.md
agitter: I had actually assumed that the pre-trained models often would come from a public repository, so if that was not clear then I agree we should make it more explicit.  It may not work all the time, but I see reusing pre-trained public models as a generally-recommended strategy, especially in image analysis (e.g. microscopy) and regulatory genomics.

The kipoi model zoo (https://doi.org/10.1101/375345) is a good source of pre-trained models for regulatory genomics.  For imaging, generic non-bio models are a common starting point.

rasbt: I think both are valid approaches. Also, they could be even combined: (1) obtain model pretrained on e.g., imagenet -> (2) fine tune on large-scale public x-ray dataset -> (3) finetune to your x-ray dataset. I don't have much practical experinence with TL though, and I am not sure what would work better in practice, i.e,. (1) -> (2) -> (3), or (1) -> (3), or (2) -> (3).

 I'd say we should definitely reference the paper and include a link to https://kipoi.org, it looks pretty useful.

Benjamin-Lee: I just pushed a change that (I hope) clarifies the distinction and applicability of pre trained models and data for pre training.

alxndrkalinin: @Benjamin-Lee thanks for addressing this! I think it's a helpful clarification.

PR #135, opened 2019-02-10
MGC author info

chevrm:

**Any more details?**


chevrm:
I think I dropped a hyphen in the yaml that's causing @khyu to not render. @Benjamin-Lee Can you take a look?

Benjamin-Lee:
Ok I think I fixed it. Thanks for pointing that out!

PR #142, opened 2019-02-22
Add explicit coverage of test-set locking (#20) to ml-concepts tip.

### LINKED ISSUE #20: Lock testing cohorts and only evaluate a trained model once.
This is probably the most occurring mistake. A lot of people test-test-test and test different models on the same testing cohorts, and then only report the best performing model. The deep learning model is, of course, over fitted. The tuning datasets can be used more frequently, but the testing should be locked and not used before the final model is identified

fmaguire:

- [X] Yes, I added myself or am already a contributor

**Any more details?**

Just add #20 more explicitly to general ML concepts to close this issue.

Fixes #20

### DIFF
--- a/content/03.ml-concepts.md
+++ b/content/03.ml-concepts.md
@@ -3,8 +3,9 @@ Deep learning is a distinct subfield of machine learning, but it is still a subf
 DL has proven to be an extremely powerful paradigm capable of outperforming “traditional” machine learning approaches in certain contexts, but it is not immune to the many limitations inherent to machine learning.
 Many best practices for machine learning also apply to deep learning.
 For instance, deep supervised learning models should be trained, tuned, and tested on non-overlapping datasets.
-Similarly, to prevent overfitting, the data used for testing should be locked and only used one-time for evaluating the very final tuned model.
-Those developing deep learning models should select data that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions.
+Similarly, to prevent overfitting, the data used for testing should be locked and only used one-time for evaluating the final model after all tuning steps were completed. 
+Using a test set more than once will lead to biased estimates of the generalization performance.
+Those developing deep learning models should also select data that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions.
 Furthermore, investigators should begin by thoroughly inspecting their data.
 When coupled with imprudence, data that is biased, skewed, or of low quality will produce models of dubious performance and limited generalizability.
 Biases in testing data can also unduly influence measures of model performance.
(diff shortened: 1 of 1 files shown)


content/03.ml-concepts.md
```diff
@@ -3,6 +3,7 @@ Deep learning is a distinct subfield of machine learning, but it is still a subf
 DL has proven to be an extremely powerful paradigm capable of outperforming “traditional” machine learning approaches in certain contexts, but it is not immune to the many limitations inherent to machine learning.
 Many best practices for machine learning also apply to deep learning.
 For instance, deep supervised learning models should be trained, tuned, and tested on non-overlapping datasets.
+Similarly, to prevent overfitting, the data used for testing should be locked and only used one-time for evaluating the very final tuned model.
```
rasbt: Very important point! Maybe we could rephrase " very final tuned model" to 

> ... the final model after all tuning steps were completed. Using a test set more than once will lead to biased estimates of the generalization performance.

fmaguire: Done

content/03.ml-concepts.md
```diff
@@ -3,7 +3,9 @@ Deep learning is a distinct subfield of machine learning, but it is still a subf
 DL has proven to be an extremely powerful paradigm capable of outperforming “traditional” machine learning approaches in certain contexts, but it is not immune to the many limitations inherent to machine learning.
 Many best practices for machine learning also apply to deep learning.
 For instance, deep supervised learning models should be trained, tuned, and tested on non-overlapping datasets.
-Those developing deep learning models should select data that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions.
+Similarly, to prevent overfitting, the data used for testing should be locked and only used one-time for evaluating the final model after all tuning steps were completed. 
+Using a test set more than once will lead to biased estimates of the generalization performance.
```
agitter: It would be nice to offer a citation here for more discussion of this issue.  I don't have any immediate suggestions though.

rasbt: It's hard to find a specific reference as this is usually already assumed in statistics and then later in ML (as independence it is one of our main assumptions for the definition of the genarlization error Expectation[L(Y,X]) but we could cite a model evaluation paper that proposes solutions (or more like practical workarounds) to the independence violation issue. A good one would be:

-  Dietterich, T. G. (1998). Approximate statistical tests for comparing supervised classification learning algorithms. Neural computation, 10(7):1895–1923. (
https://www.mitpressjournals.org/doi/pdfplus/10.1162/089976698300017197)

Not research but more of an educational write-up, I also talked about it extensively here: https://arxiv.org/abs/1811.12808




buthttps://arxiv.org/abs/1811.12808

rasbt: Do you think either of the two (or both) should be added as citations? Happy to make a PR then.

Benjamin-Lee: I'd say PR both, but especially your write-up since it's at the perfect level for our reader (looking forward to reading it myself).

rasbt: Sounds good, added in #143 !

Issue #37, opened 2018-11-14
Don't forget your ML fundamentals/Rules that apply to ML also apply to DL


Benjamin-Lee:
**Have you checked the [list of proposed rules](https://github.com/Benjamin-Lee/deep-rules/issues?q=is%3Aissue+is%3Aopen+label%3Arule) to see if the rule has already been proposed?**

- [x] Yes

**Feel free to elaborate, rant, and/or ramble.**

Maybe we could dedicate one rule to ML basics that people still don't seem to get right, which, although they are not DL specific, still apply to it. Issues #1, #19, #20, #21 all come to mind. This could possibly be rule one.

agitter:
[Machine learning applications in genetics and genomics](http://doi.org/10.1038/nrg3920) is a good reference for ML basics in biology, in addition to the manuscript in #1.

fmaguire:
One area I think would be good to add to this is more explicitly: a trained ML model itself (almost always) isn't the scientifically interesting thing in and of itself.  It's what you do with it.

The scientifically important/relevant part happens is often using your inductive ML approach and identifying where the model works well/works poorly and digging deeply into the characteristics of these to make deductive inferences and hypotheses.

This is mentioned to a degree in https://github.com/Benjamin-Lee/deep-rules/blob/master/content/11.interpretation.md

But I think it would be good to make explicit either in this tip or the intro, if anything to try and stem the tide of papers which fit a model and call it a day adding very little to the field.

PR #144, opened 2019-02-24
Minor edits to Tip 9

blengerich:

This PR includes small edits to Tip 9 intended to make discussion of prediction vs understanding more explicit, as suggested in #37.

It also might be worth considering changing the tip title to a positive phrasing. I'm not sure how to do so while maintaining distinction between Tips 9 and 8.


Benjamin-Lee:
By the way, as the primary author of a tip, please PR "yes" to accountable and add your info to the bottom of `metadata.yaml` so we can get your name on the paper!

PR #147, opened 2019-02-26
Conclusion first draft

Benjamin-Lee:

**Any more details?**

### DIFF
--- a/content/13.conclusion.md
+++ b/content/13.conclusion.md
@@ -1,1 +1,6 @@
 ## Conclusion {#conclusion}
+
+Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in an variety of tasks while generating biological insights that may be applied back to fundamental research. 
+To realize this potential, the use of deep learning as a research tool must be approached as any other tool would be: thoughtfully, systematically, and [**TODO**: fill in an adverb here. Ideas: deliberately, ethically, carefully, reproducibly].
+We hope that the tips we present will serve as a starting point for the discussion of best practices for deep learning as they apply to biology, not as an ending point.
+As the field of deep learning matures, with its theoretical underpinnings more fully understood, the best practices for its use will surely evolve as well.
--- a/content/13.conclusion.md
+++ b/content/13.conclusion.md
@@ -1,6 +1,6 @@
 ## Conclusion {#conclusion}
 
 Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in a variety of tasks while generating biological insights that may be applied back to fundamental research. 
-To realize this potential, the use of deep learning as a research tool must be approached as any other tool would be: thoughtfully, systematically, and [**TODO**: fill in an adverb here. Ideas: deliberately, ethically, carefully, reproducibly].
+To realize this potential, the use of deep learning as a research tool must be approached as any other tool would be: scientifically and thoughtfully.
 We hope that the tips we present here will serve as a starting point for the discussion of best practices for deep learning as they apply to biology, not as an ending point.
 As the field of deep learning matures, with its theoretical underpinnings more fully understood, the best practices for its use will surely evolve as well.
--- a/content/13.conclusion.md
+++ b/content/13.conclusion.md
@@ -1,6 +1,6 @@
 ## Conclusion {#conclusion}
 
-Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in an variety of tasks while generating biological insights that may be applied back to fundamental research. 
+Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in a variety of tasks while generating biological insights that may be applied back to fundamental research. 
 To realize this potential, the use of deep learning as a research tool must be approached as any other tool would be: thoughtfully, systematically, and [**TODO**: fill in an adverb here. Ideas: deliberately, ethically, carefully, reproducibly].
-We hope that the tips we present will serve as a starting point for the discussion of best practices for deep learning as they apply to biology, not as an ending point.
+We hope that the tips we present here will serve as a starting point for the discussion of best practices for deep learning as they apply to biology, not as an ending point.
 As the field of deep learning matures, with its theoretical underpinnings more fully understood, the best practices for its use will surely evolve as well.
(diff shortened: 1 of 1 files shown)


content/13.conclusion.md
```diff
@@ -1 +1,6 @@
 ## Conclusion {#conclusion}
+
+Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in an variety of tasks while generating biological insights that may be applied back to fundamental research.
```
AlexanderTitus: "...algorithms in a variety..."

content/13.conclusion.md
```diff
@@ -1 +1,6 @@
 ## Conclusion {#conclusion}
+
+Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in an variety of tasks while generating biological insights that may be applied back to fundamental research. 
+To realize this potential, the use of deep learning as a research tool must be approached as any other tool would be: thoughtfully, systematically, and [**TODO**: fill in an adverb here. Ideas: deliberately, ethically, carefully, reproducibly].
```
AlexanderTitus: thoughtfully, logically, and systematically?

content/13.conclusion.md
```diff
@@ -1 +1,6 @@
...
+We hope that the tips we present will serve as a starting point for the discussion of best practices for deep learning as they apply to biology, not as an ending point.
```
AlexanderTitus: "...we present here will..."

content/13.conclusion.md
cgreene: I like the use of "scientifically" somewhere. I like it more than systematically :)

Benjamin-Lee: Ok what about: "scientifically, thoughtfully, and [adverb]".  "Logically" and "deliberately" seem to be the covered by "thoughtfully" and "reproducibly" is encompassed by "scientifically".

cgreene: What about thoughtfully and scientifically and leave it at that? Good things come in threes, but in this case since we can't figure out the third maybe going with two we're confident in is better 😁

AlexanderTitus: Agreed, lets keep it simple. I like using the two.

content/13.conclusion.md
```diff
@@ -1 +1,6 @@
 ## Conclusion {#conclusion}
+
+Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in a variety of tasks while generating biological insights that may be applied back to fundamental research.
```
AlexanderTitus: This is a long sentence. Can we break up?

content/13.conclusion.md
```diff
@@ -1 +1,6 @@
 ## Conclusion {#conclusion}
+
+Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in a variety of tasks while generating biological insights that may be applied back to fundamental research. 
+To realize this potential, the use of deep learning as a research tool must be approached as any other tool would be: scientifically and thoughtfully.
```
AlexanderTitus: I like the order thoughtfully and scientifically.

content/13.conclusion.md
Benjamin-Lee: Done!

content/13.conclusion.md
```diff
@@ -1 +1,7 @@
 ## Conclusion {#conclusion}
+
+Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in a variety of tasks.
+Beyond simply achieving performance, deep learning has the potential to generate novel biological insights that may then be applied back to fundamental research.
```
rasbt: I don't think "simply achieving performance" is correct -- I don't think someone/something can "achieve performance." You meant "good predictive performance?"

content/13.conclusion.md
```diff
@@ -1 +1,7 @@
...
+To realize this potential, the use of deep learning as a research tool must be approached as any other tool would be: scientifically and thoughtfully.
```
rasbt: Nothing wrong with it, but we could simplify this to

> To realize this potential, the use of deep learning as a research tool must be approached scientifically and thoughtfully.

content/13.conclusion.md
```diff
@@ -1 +1,7 @@
...
+We hope that the tips we present here will serve as a starting point for the discussion of best practices for deep learning as they apply to biology, not as an ending point.
```
rasbt: Also here, we can simplify for readability:

"We hope that the tips we present here " -> "We hope that our tips"

content/13.conclusion.md
```diff
@@ -1 +1,7 @@
...
+As the field of deep learning matures, with its theoretical underpinnings more fully understood, the best practices for its use will surely evolve as well.
```
rasbt: Again, nothing wrong with it, but very speculative. I am not sure if we need that sentence.

content/13.conclusion.md
Benjamin-Lee: While simpler, I think that the proposed alternative does not capture the key phrase "any other tool". However, the current line 5 might not get what I am trying to say across either.

To make what I was trying to get at concrete, most biologists don't anthropomorphize PCR or see it as black magic; they simply use it as a powerful but delicate tool. In an ideal world, I'd like to see DL used similarly to PCR in that respect.

rasbt: I see. Maybe adding a concrete example would help in that case then? I was more thinking about tools as "programming" or things like that (in that context, using it properly in terms of documenting, making sure the outcome is understandable and reproducible etc.). Regarding PCR, your comment makes sense then, but since the audience will be mostly biologists, maybe a concrete example (like the PCR one) that everyone can relate to would be nice? (Esp. because I think our article is very vague in general ... readers might appreciate some examples)

content/13.conclusion.md
```diff
@@ -1,7 +1,6 @@
 ## Conclusion {#conclusion}
 
 Deep learning techniques have the potential for wide use in biology, meeting or exceeding the performance of both humans and the current state-of-the art algorithms in a variety of tasks.
-Beyond simply achieving performance, deep learning has the potential to generate novel biological insights that may then be applied back to fundamental research.
+Beyond simply achieving good predictive performance, deep learning has the potential to generate novel biological insights that may then be applied back to fundamental research.
```
rasbt: > biological insights that may then be applied back to fundamental research.

Another little suggestion as this sounds a bit awkward: how about changing this to 

> biological insights that can assist the progress of fundamental research.

PR #157, opened 2019-03-05
Some more citations and fixes to complexities tip

Benjamin-Lee:

**Any more details?**

### DIFF
--- a/content/05.dl-complexities.md
+++ b/content/05.dl-complexities.md
@@ -6,7 +6,7 @@ To get good results, you must expect to train many networks with a range of diff
 Despite improved framework ease-of-use and on-demand cloud computing resources this means DL can be very demanding, often requiring significant infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740].
 The experimentation inherent to DL is often noisy (requiring repetition) and represents a significant organizational challenge.
 All code, random seeds, parameters, and results must be carefully corralled using general good coding practices (for example, version control [@doi:10.1371/journal.pcbi.1004947], continuous integration etc.) in order to be effective and interpretable.
-This organization is also key to being able to efficiently share and reproduce your work as well as to update your model as new data becomes available [@doi:10.1371/journal.pcbi.1003285; @arxiv:1810.08055].
+This organization is also key to being able to efficiently share and reproduce your work [@doi:10.1371/journal.pcbi.1003285; @arxiv:1810.08055] as well as to update your model as new data becomes available.
 
 One specific reproducibility pitfall that is often missed in deep learning applications is the default use of non-deterministic algorithms by CUDA/CuDNN backends when training models on GPUs.
 Making this process reproducible is distinct from setting random seeds, which will primarily affect pseudorandom deterministic procedures such as shuffling and initialization, and requires explicitly specifying the use of deterministic algorithms in your DL library [@url:https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#reproducibility]. 
--- a/content/05.dl-complexities.md
+++ b/content/05.dl-complexities.md
@@ -3,7 +3,7 @@
 Correctly training deep neural networks is a non-trivial process.
 There are many different options and potential pitfalls at every stage.
 To get good results, you must expect to train many networks with a range of different parameter and hyperparameter settings.
-Despite improved framework ease-of-use and on-demand cloud computing resources this means DL can be very demanding, often requiring significant infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740].
+Deep learning can be very demanding, often requiring significant computing infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740].
 The experimentation inherent to DL is often noisy (requiring repetition) and represents a significant organizational challenge.
 All code, random seeds, parameters, and results must be carefully corralled using general good coding practices (for example, version control [@doi:10.1371/journal.pcbi.1004947], continuous integration etc.) in order to be effective and interpretable.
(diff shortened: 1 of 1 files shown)


content/05.dl-complexities.md
```diff
@@ -2,13 +2,13 @@
 
 Correctly training deep neural networks is a non-trivial process.
 There are many different options and potential pitfalls at every stage.
-To get good results you have to expect to train many networks with a range of different parameter and hyperparameter settings.
+To get good results, you must expect to train many networks with a range of different parameter and hyperparameter settings.
 Despite improved framework ease-of-use and on-demand cloud computing resources this means DL can be very demanding, often requiring significant infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740].
```
Benjamin-Lee: This line doesn't seem to correct, but I'm not quite sure what framework ease-of-use and DL infrastructure requirements have in common. Can we make this line clearer?

content/05.dl-complexities.md
```diff
@@ -2,13 +2,13 @@
...
 The experimentation inherent to DL is often noisy (requiring repetition) and represents a significant organizational challenge.
-All code, random seeds, parameters, and results must be carefully corralled using good coding practices (for example, version control, continuous integration etc.) in order to be effective and interpretable.
-This organization is also key to being able to efficiently share your work and to update your model as new data becomes available. 
+All code, random seeds, parameters, and results must be carefully corralled using general good coding practices (for example, version control [@doi:10.1371/journal.pcbi.1004947], continuous integration etc.) in order to be effective and interpretable.
+This organization is also key to being able to efficiently share and reproduce your work as well as to update your model as new data becomes available [@doi:10.1371/journal.pcbi.1003285; @arxiv:1810.08055].
```
agitter: Perhaps move these citations to follow `reproduce your work`?  They have more to do with that than updating models.

content/05.dl-complexities.md
Benjamin-Lee: To make this syntactically (but not necessarily semantically) correct, how about:

"Despite improved framework ease-of-use and on-demand cloud computing resources, DL can be very demanding, often requiring significant infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740]."

agitter: Should we start the sentence with "DL can be very demanding" and move the initial clause elsewhere?  I think the original idea was that even though it is easier to implement and run DL models, that doesn't change the local or cloud computational demands.

Benjamin-Lee: How about:
> DL [training?] can be very demanding, often requiring significant computing infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740].

agitter: Yes, I like that.  I lean toward excluding "training" in case that implies training a single model is hard.  Often the real demand is in hyperparameter search and training many models.

Benjamin-Lee: What about "model optimization" or "hyperparameter search" instead of training?

Benjamin-Lee: You're right; there's no need for that word since the last sentence makes it clear what we're talking about

content/05.dl-complexities.md
```diff
@@ -2,13 +2,13 @@
 
 Correctly training deep neural networks is a non-trivial process.
 There are many different options and potential pitfalls at every stage.
-To get good results you have to expect to train many networks with a range of different parameter and hyperparameter settings.
-Despite improved framework ease-of-use and on-demand cloud computing resources this means DL can be very demanding, often requiring significant infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740].
+To get good results, you must expect to train many networks with a range of different parameter and hyperparameter settings.
+Deep learning can be very demanding, often requiring significant computing infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740].
```
rasbt: > significant computing infrastructure

maybe "extensive" instead of "significant?"

content/05.dl-complexities.md
```diff
@@ -2,13 +2,13 @@
...
-All code, random seeds, parameters, and results must be carefully corralled using good coding practices (for example, version control, continuous integration etc.) in order to be effective and interpretable.
-This organization is also key to being able to efficiently share your work and to update your model as new data becomes available. 
+All code, random seeds, parameters, and results must be carefully corralled using general good coding practices (for example, version control [@doi:10.1371/journal.pcbi.1004947], continuous integration etc.) in order to be effective and interpretable.
+This organization is also key to being able to efficiently share and reproduce your work [@doi:10.1371/journal.pcbi.1003285; @arxiv:1810.08055] as well as to update your model as new data becomes available.
 
-One specific reproducibility pitfall that is often missed is the default use of non-deterministic algorithms by CUDA/CuDNN backends when training on GPUs.
+One specific reproducibility pitfall that is often missed in deep learning applications is the default use of non-deterministic algorithms by CUDA/CuDNN backends when training models on GPUs.
```
rasbt: technically, these algorithms are also non-deterministic during prediction (after training). It's just not as noticeable. I would rephrase to 

> ... by CUDA/CuDNN backends when ~~training models on~~ [using] GPUs.

content/05.dl-complexities.md
```diff
@@ -2,13 +2,13 @@
...
 Making this process reproducible is distinct from setting random seeds, which will primarily affect pseudorandom deterministic procedures such as shuffling and initialization, and requires explicitly specifying the use of deterministic algorithms in your DL library [@url:https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#reproducibility]. 
 
 Similar to [Tip 4](#baselines), try to start with a relatively smaller network and increase the size and complexity as needed to prevent wasting time and resources.
```
rasbt: > ...relatively ~~smaller~~ [small] network a ...

Benjamin-Lee:
@rasbt Done!

PR #162, opened 2019-03-11
Some tweaks to baselines/tip 2. Get rid of informal language in first sentence

pstew:

**Any more details?**

### DIFF
--- a/content/04.baselines.md
+++ b/content/04.baselines.md
@@ -1,7 +1,7 @@
 ## Tip 2: Use traditional methods to establish performance baselines {#baselines}
 
 It is important to implement a simple model to establish an adequate performance baseline, since it is easy to fall into the trap of an unnecessarily convoluted analysis.
-For example, a researcher could first a build a multinomial logistic regression or random forest model using the same software framework planned for DL to first evaluate the classification performance.
+For example, a researcher could first a build a logistic regression model using the same software framework planned for DL to first evaluate the classification performance.
 This approach will help researchers with assessing the complexity of the task at hand and debugging more complex DL architectures.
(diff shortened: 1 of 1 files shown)


content/04.baselines.md
```diff
@@ -1,14 +1,14 @@
 ## Tip 2: Use traditional methods to establish performance baselines {#baselines}
 
-Before diving into a fancy thousand-layer neural network, always implement at least a simple model to establish an adequate performance baseline.
-For example, researchers can build multinomial logistic regression or random forest models using the same software framework that is being used for DL and evaluate its classification performance.
+It is important to implement a simple model to establish an adequate performance baseline, since it is easy to fall into the trap of an unnecessarily convoluted analysis.
+For example, a researcher could first a build a multinomial logistic regression or random forest model using the same software framework planned for DL to first evaluate the classification performance.
```
agitter: This isn't your edit, but do popular DL frameworks support random forest?  That recommendation seemed out of place to me in the initial draft.  Logistic regression makes sense.

Benjamin-Lee: Not that I know of. I didn't write this but I was thinking of the Scikit-Learn API for fitting simple models, [AutoML models](https://epistasislab.github.io/tpot/), and DL models (via Keras)

pstew: Not to my knowledge.

rasbt: Also, RF is probably not feasible on most DL datasets (i.e., thinking of image classification with >500x500 pixels)

agitter: I propose changing to `a researcher could first a build a logistic regression model using the same software framework`.  The `multinomial` part will depend on the data type.

rasbt: sounds good. I don't think we need to say "multinomial" explicitly.

PR #168, opened 2019-03-13
Tip 1 edits

rasbt:

**Any more details?**

Some edits for Tip 1

### DIFF
--- a/content/03.ml-concepts.md
+++ b/content/03.ml-concepts.md
@@ -8,7 +8,7 @@ However, the ability of deep learning to perform implicit feature extraction can
 For instance, those developing deep learning models not only need to ensure that the data contains information relevant to the problem at hand but is also free of systematic biases.
 As deep learning models are capable of extracting and learning from complex patterns in a training dataset, DL models are also susceptible to systematic noise and non-salient information, which can not only hamper performance but also lead to spurious conclusions.
 
-Since deep learning still remains a subfield of machine learning, the majority of best practices for machine learning still apply to deep learning.
+Since deep learning is still a subfield of machine learning, the majority of best practices for machine learning still apply to deep learning.
 Like all computational methods, experiments involving deep learning should be conducted in a systematic manner and accompanied by reproducible instructions as well as rigorous model evaluation.
 For instance, biases in testing data can unduly influence measures of model performance, and it may be difficult to identify confounders from the model directly.
 Investigators should consider the extent to which the outcome of interest is likely to be predictable from the input data and begin by thoroughly inspecting the input data.
--- a/content/03.ml-concepts.md
+++ b/content/03.ml-concepts.md
@@ -1,24 +1,28 @@
 ## Tip 1: Concepts that apply to machine learning also apply to deep learning {#concepts}
-Deep learning is a distinct subfield of machine learning, but it is still a subfield.
-DL has proven to be an extremely powerful paradigm capable of outperforming “traditional” machine learning approaches in certain contexts, but it is not immune to the many limitations inherent to machine learning.
-Many best practices for machine learning also apply to deep learning.
-Like all computational methods, deep learning should be applied in a systematic manner that is reproducible and rigorously tested.
 
-Those developing deep learning models should select data that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions.
-Biases in testing data can also unduly influence measures of model performance, and it may be difficult to directly identify confounders from the model.
-Investigators should consider the extent to which the outcome of interest is likely to be predictable from the input data and begin by throughly inspecting the input data.
-Suppose that there are robust heritability estimates for a phenotype that suggest that the genetic contribution is modest but a deep learning model predicts the phenotype with very high accuracy.
+While deep learning has inarguably surpassed general machine learning for artificial intelligence and predictive modeling applications in terms of public awareness, deep learning still remains a subfield of machine learning.
+DL has proven to be an extremely powerful paradigm capable of outperforming "traditional" or "conventional" machine learning approaches. This is partly owed to the fact that DL algorithms are less dependent on manual feature extraction and are capable of achieving high predictive performance when operating on raw-data, such as images and text. directly.
+In contrast, traditional machine learning approaches often require manual feature extraction to produce acceptable results.
+However, the majority of best practices for machine learning still apply to deep learning.
+Like all computational methods, experiments involving deep learning should be conducted in a systematic manner and accompanied by reproducible instructions as well as rigorous model evaluation.
+
+The ability of deep learning to perform implicit feature extraction can be regarded as a double-edged sword.
+For instance, those developing deep learning models not only need to ensure that the data contains information relevant to the problem at hand but is also free of systematic biases.
+As deep learning models are capable of extracting and learning from complex patterns in a training dataset, DL models are also susceptible to systematic noise and non-salient information, which can not only hamper performance but also lead to spurious conclusions.
+Biases in testing data can also unduly influence measures of model performance, and it may be difficult to identify confounders from the model directly.
+Investigators should consider the extent to which the outcome of interest is likely to be predictable from the input data and begin by thoroughly inspecting the input data.
+Suppose that there are robust heritability estimates for a phenotype that suggest that the genetic contribution is modest, but a deep learning model predicts the phenotype with very high accuracy.
 The model may be capturing signal unrelated to genetic mechanisms underlying the phenotype.
 In this case, a possible explanation is that people with similar genetic markers may have shared exposures.
-This is something that researchers should probe before reporting unrealistic accuracy measures.
-A similar situation can arise with tasks for which inter-rater reliability is modest but deep learning models produce very high accuracies.
+This is something that researchers should probe before reporting unrealistically high prediction performances.
+A similar situation can arise with tasks for which inter-rater reliability is modest but deep learning models produce very high prediction accuracies.
 When coupled with imprudence, data that is confounded, biased, skewed, or of low quality will produce models of dubious performance and limited generalizability.
 
 Using a test set more than once will lead to biased estimates of the generalization performance  [@arxiv:1811.12808; @doi:10.1162/089976698300017197].
-Deep supervised learning models should be trained, tuned, and tested on non-overlapping datasets.
-The data used for testing should be locked and only used one-time for evaluating the final model after all tuning steps are completed.
-Also, many conventional metrics for classification (e.g. area under the receiver operating characteristic curve or AUROC) have limited utility in cases of extreme class imbalance [@pmid:25738806].
+Consequently, supervised deep learning models should be trained, tuned, and tested on non-overlapping datasets.
+The data used for testing should be withheld during model tuning and refinement; test data should only be used one-time for evaluating the final model after all tuning steps are completed.
+Also, many conventional metrics for classification (e.g., area under the receiver operating characteristic curve or AUROC) have limited utility in cases of extreme class imbalance [@pmid:25738806].
 Model performance should be evaluated with a carefully-picked panel of relevant metrics that make minimal assumptions about the composition of the testing data [@doi:10.1021/acs.molpharmaceut.7b00578], with particular consideration given to metrics that are most directly applicable to the task at hand.
 
-Extreme cases warrant testing the robustness of the model and metrics on simulated data for which the ground truth is known.
-Said simulations can be used to verify the correctness of the model’s implementation as well.
+Extreme cases warrant testing the robustness of the model and metrics on simulated data for which the ground truth (i.e., the information to be predicted) is known.
+Said simulations can be used to verify the technical correctness of a model's implementation as well.
--- a/content/03.ml-concepts.md
+++ b/content/03.ml-concepts.md
@@ -1,15 +1,16 @@
-## Tip 1: Concepts that apply to machine learning also apply to deep learning {#concepts}
+## Tip 1: Successful deep learning applications require best practices for machine learning and model evaluation {#concepts}
 
-While deep learning has inarguably surpassed general machine learning for artificial intelligence and predictive modeling applications in terms of public awareness, deep learning still remains a subfield of machine learning.
-DL has proven to be an extremely powerful paradigm capable of outperforming "traditional" or "conventional" machine learning approaches. This is partly owed to the fact that DL algorithms are less dependent on manual feature extraction and are capable of achieving high predictive performance when operating on raw-data, such as images and text. directly.
-In contrast, traditional machine learning approaches often require manual feature extraction to produce acceptable results.
-However, the majority of best practices for machine learning still apply to deep learning.
-Like all computational methods, experiments involving deep learning should be conducted in a systematic manner and accompanied by reproducible instructions as well as rigorous model evaluation.
 
-The ability of deep learning to perform implicit feature extraction can be regarded as a double-edged sword.
+In recent years, deep learning has proven to be an extremely powerful paradigm capable of outperforming "traditional" or "conventional" machine learning approaches.
+The recent success behind applying deep learning is partly owed to the fact that DL algorithms are less dependent on manual feature extraction and are capable of achieving high predictive performance when operating on raw-data, such as images and text directly.
+In contrast, traditional machine learning approaches often require manual feature extraction to produce acceptable results. 
+However, the ability of deep learning to perform implicit feature extraction can be regarded as a double-edged sword.
 For instance, those developing deep learning models not only need to ensure that the data contains information relevant to the problem at hand but is also free of systematic biases.
 As deep learning models are capable of extracting and learning from complex patterns in a training dataset, DL models are also susceptible to systematic noise and non-salient information, which can not only hamper performance but also lead to spurious conclusions.
-Biases in testing data can also unduly influence measures of model performance, and it may be difficult to identify confounders from the model directly.
+
+Since deep learning still remains a subfield of machine learning, the majority of best practices for machine learning still apply to deep learning.
+Like all computational methods, experiments involving deep learning should be conducted in a systematic manner and accompanied by reproducible instructions as well as rigorous model evaluation.
+For instance, biases in testing data can unduly influence measures of model performance, and it may be difficult to identify confounders from the model directly.
 Investigators should consider the extent to which the outcome of interest is likely to be predictable from the input data and begin by thoroughly inspecting the input data.
 Suppose that there are robust heritability estimates for a phenotype that suggest that the genetic contribution is modest, but a deep learning model predicts the phenotype with very high accuracy.
 The model may be capturing signal unrelated to genetic mechanisms underlying the phenotype.
--- a/content/03.ml-concepts.md
+++ b/content/03.ml-concepts.md
@@ -1,4 +1,4 @@
-## Tip 1: Successful deep learning applications require best practices for machine learning and model evaluation {#concepts}
+## Tip 1: Adhere to machine learning best practices for rigorous model evaluation {#concepts}
 
 
 In recent years, deep learning has proven to be an extremely powerful paradigm capable of outperforming "traditional" or "conventional" machine learning approaches.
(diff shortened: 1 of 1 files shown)


content/03.ml-concepts.md
```diff
@@ -1,24 +1,28 @@
...
 When coupled with imprudence, data that is confounded, biased, skewed, or of low quality will produce models of dubious performance and limited generalizability.
 
 Using a test set more than once will lead to biased estimates of the generalization performance  [@arxiv:1811.12808; @doi:10.1162/089976698300017197].
-Deep supervised learning models should be trained, tuned, and tested on non-overlapping datasets.
-The data used for testing should be locked and only used one-time for evaluating the final model after all tuning steps are completed.
-Also, many conventional metrics for classification (e.g. area under the receiver operating characteristic curve or AUROC) have limited utility in cases of extreme class imbalance [@pmid:25738806].
+Consequently, supervised deep learning models should be trained, tuned, and tested on non-overlapping datasets.
+The data used for testing should be withheld during model tuning and refinement; test data should only be used one-time for evaluating the final model after all tuning steps are completed.
```
evancofer: This seems verbose and a little confusing. It makes one wonder if "data used for testing" and "test data" are different.

content/03.ml-concepts.md
```diff
@@ -1,24 +1,28 @@
...
-Many best practices for machine learning also apply to deep learning.
-Like all computational methods, deep learning should be applied in a systematic manner that is reproducible and rigorously tested.
 
-Those developing deep learning models should select data that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions.
-Biases in testing data can also unduly influence measures of model performance, and it may be difficult to directly identify confounders from the model.
-Investigators should consider the extent to which the outcome of interest is likely to be predictable from the input data and begin by throughly inspecting the input data.
-Suppose that there are robust heritability estimates for a phenotype that suggest that the genetic contribution is modest but a deep learning model predicts the phenotype with very high accuracy.
+While deep learning has inarguably surpassed general machine learning for artificial intelligence and predictive modeling applications in terms of public awareness, deep learning still remains a subfield of machine learning.
```
evancofer: The aside about public awareness almost comes across as a little sarcastic/snide, and I don't think it really fits with the rest of the paper's tone.

content/03.ml-concepts.md
```diff
@@ -1,24 +1,28 @@
...
+DL has proven to be an extremely powerful paradigm capable of outperforming "traditional" or "conventional" machine learning approaches. This is partly owed to the fact that DL algorithms are less dependent on manual feature extraction and are capable of achieving high predictive performance when operating on raw-data, such as images and text. directly.
```
evancofer: I think we should retain the caveat about "certain contexts", since there are still places where deep learning doesn't make sense (e.g. small sample size problems are probably still best approached with strictly Bayesian methods). 

This also needs to be split into one line per sentence.

content/03.ml-concepts.md
```diff
@@ -1,24 +1,28 @@
...
+In contrast, traditional machine learning approaches often require manual feature extraction to produce acceptable results.
+However, the majority of best practices for machine learning still apply to deep learning.
+Like all computational methods, experiments involving deep learning should be conducted in a systematic manner and accompanied by reproducible instructions as well as rigorous model evaluation.
+
+The ability of deep learning to perform implicit feature extraction can be regarded as a double-edged sword.
+For instance, those developing deep learning models not only need to ensure that the data contains information relevant to the problem at hand but is also free of systematic biases.
+As deep learning models are capable of extracting and learning from complex patterns in a training dataset, DL models are also susceptible to systematic noise and non-salient information, which can not only hamper performance but also lead to spurious conclusions.
```
evancofer: I would argue that many ML methods are capable of extracting near-arbitrary patterns from the data, but DL introduces additional complexity because it can scale so well. With DL you end up with data of such high dimensionality that it is nearly impossible to examine the impact of each component of the input. Furthermore, the size of datasets used for DL are often quite large, and a thorough examination of their biases (which is generally domain-specific) may take much longer than selecting and developing the model.

content/03.ml-concepts.md
```diff
@@ -1,24 +1,28 @@
...
-Suppose that there are robust heritability estimates for a phenotype that suggest that the genetic contribution is modest but a deep learning model predicts the phenotype with very high accuracy.
+While deep learning has inarguably surpassed general machine learning for artificial intelligence and predictive modeling applications in terms of public awareness, deep learning still remains a subfield of machine learning.
+DL has proven to be an extremely powerful paradigm capable of outperforming "traditional" or "conventional" machine learning approaches. This is partly owed to the fact that DL algorithms are less dependent on manual feature extraction and are capable of achieving high predictive performance when operating on raw-data, such as images and text. directly.
+In contrast, traditional machine learning approaches often require manual feature extraction to produce acceptable results.
+However, the majority of best practices for machine learning still apply to deep learning.
+Like all computational methods, experiments involving deep learning should be conducted in a systematic manner and accompanied by reproducible instructions as well as rigorous model evaluation.
```
evancofer: If we are going to have this large discussion of automatic feature extraction, I think this mention of reproducible research should go before it. The mention of reproducible computational research is in line with the notion that DL is a subset of ML/computation, while the discussion of feature extraction is a more ML/DL-specific one. I think this change will help this section to flow from a general case to more ML/DL-specific scenarios, and allow the discussion of feature extraction to move more naturally into the following discussion of biases in data and evaluation.

content/03.ml-concepts.md
Benjamin-Lee: I, for one, jumped straight to deep learning without any knowledge of ML and it was a bit of a train wreck, so I do think that this sentence has value to the complete newcomer. Regardless, if this sentence came off as snide/sarcastic to you (it didn't to me), it is likely that it will to others as well so we should change it.

rasbt: Thanks for the note! Although, I unfortunately have a tendency for sarcasm, I didn't intend to be sarcastic here ... But I agree that we should change it ... Will think about an alternative way to say it

content/03.ml-concepts.md
rasbt: > I would argue that many ML methods are capable of extracting near-arbitrary patterns from the data, but DL introduces additional complexity because it can scale so well. 

Hm, but I'd say the majority of methods, like Logistic regression, linear SVMs, decision trees,  etc. can't, though. But yeah, but kernel SVMs with non-polynomial kernels do e.g., some sort of implicit feature extraction ...

> DL introduces additional complexity because it can scale so well. 

Hm, do you mean scaling with regard to a large numbers of features? In my opinion, that's only because of the implicit feature extraction. E.g., in convnets you have the convolutional kernels that combine features (vs. doing it manually) and you also have weight sharing to actually AVOID operating on a large number of features for making predictions. Usually, the (final) prediction layer is only ~100-2000 features large; I see everything that comes before that as implicit feature extraction layers.

You can actually empirically show that by e.g., implementing an autoencoder and then training a simple logistic regression classifier on the latent space. It will usually perform reasonably well on this compressed feature space (thanks to the feature extraction) whereas it would perform much worse if you'd trained it on the high-dimensional input layer directly.

content/03.ml-concepts.md
rasbt: That's a good point! I re-organized this a bit. Actually, having the intro in mind, I think saying "DL is a subfield of machine learning" again and again may be a bit too preachery.  I think what we really just want to say is that careful model evaluation is important. Maybe see whether the reorganization makes sense to you ...

content/03.ml-concepts.md
evancofer: My guess is you won't be applying linear SVMs to the same problems that you will use CNNs for (: I agree that feature extraction is probably why NNs can succeed on some of these datasets, but I think the high dimensionality of the data at hand may be the real "core" of the problem. I guess I am just hesitant to make non-DL methods sound "safe", when they may be just as uninterpretable when applied to the same high dimensional data :/

content/03.ml-concepts.md
```diff
@@ -1,16 +1,16 @@
...
+In contrast, traditional machine learning approaches often require manual feature extraction to produce acceptable results. 
+However, the ability of deep learning to perform implicit feature extraction can be regarded as a double-edged sword.
 For instance, those developing deep learning models not only need to ensure that the data contains information relevant to the problem at hand but is also free of systematic biases.
 As deep learning models are capable of extracting and learning from complex patterns in a training dataset, DL models are also susceptible to systematic noise and non-salient information, which can not only hamper performance but also lead to spurious conclusions.
-Biases in testing data can also unduly influence measures of model performance, and it may be difficult to identify confounders from the model directly.
+
+Since deep learning still remains a subfield of machine learning, the majority of best practices for machine learning still apply to deep learning.
```
evancofer: I think that `still remains a subfield` sounds really wordy and kind of odd here. I would change it to `Since deep learning is still a subfield`.

content/03.ml-concepts.md
rasbt: Oh I see, I think I misunderstood your concern. I was thinking about it like this

- traditional: image dataset -> extract landmarks (e.g., facial landmarks) -> apply RF/SVM
- DL: image dataset -> apply convnet (let it take care of feature extraction)

If you apply a regular MLP to large-feature sets, you would have the same problem as you'd have when applying e.g. SVM -- but then, I think just using a multilayer perceptron on tabular data wouldn't be really deep learning (I don't think though that someone would apply these methods to raw data though).

content/03.ml-concepts.md
rasbt: totally agree!

content/03.ml-concepts.md
```diff
@@ -1,24 +1,29 @@
-## Tip 1: Concepts that apply to machine learning also apply to deep learning {#concepts}
-Deep learning is a distinct subfield of machine learning, but it is still a subfield.
-DL has proven to be an extremely powerful paradigm capable of outperforming “traditional” machine learning approaches in certain contexts, but it is not immune to the many limitations inherent to machine learning.
-Many best practices for machine learning also apply to deep learning.
-Like all computational methods, deep learning should be applied in a systematic manner that is reproducible and rigorously tested.
+## Tip 1: Successful deep learning applications require best practices for machine learning and model evaluation {#concepts}
```
Benjamin-Lee: Can we make the title more "punchy"?  Or at least an imperative sentence with an implied "you" as the subject so that it matches all of the other tip titles?

Something like: 
> Follow best practices for machine learning and model evaluation

rasbt: I am not a big fan of the imperative style because it sounds like orders or commandments, but, yeah, we should be consistent! How about

> Use Machine Learning Best Practices for Model Evaluation

we could also use 

> Use Machine Learning Best Practices for Rigorous Model Evaluation

but "rigorous" has a particular meaning in science so I'd prefer the first version.

rasbt: ok changed it now

content/03.ml-concepts.md
rasbt: I think the concern should be covered in the sentence:

> In contrast, traditional machine learning approaches often require manual feature extraction to produce acceptable results.

content/03.ml-concepts.md
rasbt: I think the issue should have disappeared after rephrasing. I think the main point we want to get at here is that as DL models usually have a larger capacity and on larger feature sets (usually raw data instead of manually curated features), using machine learning best practices for model evaluation is ever so important (we list various caveats). Regarding which approach is better for a given dataset etc, I think that's covered in different tips, incl. the tip on baselines and the one on overfitting.

rasbt:
Alright, I think it'd be best if someone could take a look at it again after all that rephrasing. I think the main point here is that using machine learning best practices (which also apply to DL as DL is a subfield of ML) is super important, esp. for model evaluation. (we probably have to be careful that we don't cram too much into this one single tip).

content/03.ml-concepts.md
```diff
@@ -1,24 +1,29 @@
...
-Many best practices for machine learning also apply to deep learning.
-Like all computational methods, deep learning should be applied in a systematic manner that is reproducible and rigorously tested.
+## Tip 1: Adhere to machine learning best practices for rigorous model evaluation {#concepts}
 
-Those developing deep learning models should select data that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions.
-Biases in testing data can also unduly influence measures of model performance, and it may be difficult to directly identify confounders from the model.
-Investigators should consider the extent to which the outcome of interest is likely to be predictable from the input data and begin by throughly inspecting the input data.
-Suppose that there are robust heritability estimates for a phenotype that suggest that the genetic contribution is modest but a deep learning model predicts the phenotype with very high accuracy.
+
+In recent years, deep learning has proven to be an extremely powerful paradigm capable of outperforming "traditional" or "conventional" machine learning approaches.
```
evancofer: This paragraph seems to focus entirely on defining deep learning and setting it apart from conventional means. As such, it seems like it should go in the introduction, but not in the first tip. Moving it there might give us more space to outline key best practices in ML/DL here.

content/03.ml-concepts.md
```diff
@@ -1,24 +1,29 @@
...
+In contrast, traditional machine learning approaches often require manual feature extraction to produce acceptable results. 
+However, the ability of deep learning to perform implicit feature extraction can be regarded as a double-edged sword.
+For instance, those developing deep learning models not only need to ensure that the data contains information relevant to the problem at hand but is also free of systematic biases.
+As deep learning models are capable of extracting and learning from complex patterns in a training dataset, DL models are also very susceptible to systematic noise and non-salient information, which can not only hamper performance but also lead to spurious conclusions, making proper model evaluation an ever so important topic.
+
+While deep learning models require careful evaluation, good news is that since deep learning is still a subfield of machine learning, the majority of best practices that were developed for machine learning over the last decades also apply to DL.
```
evancofer: I would move this idea up earlier in the tip. The impetus of this tip is that "deep learning is still a subfield of machine learning", and IMO that should established at the start of tip 1. This sentence also has some grammatical errors.

content/03.ml-concepts.md
```diff
@@ -1,24 +1,29 @@
...
 
-Those developing deep learning models should select data that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions.
-Biases in testing data can also unduly influence measures of model performance, and it may be difficult to directly identify confounders from the model.
-Investigators should consider the extent to which the outcome of interest is likely to be predictable from the input data and begin by throughly inspecting the input data.
-Suppose that there are robust heritability estimates for a phenotype that suggest that the genetic contribution is modest but a deep learning model predicts the phenotype with very high accuracy.
+
+In recent years, deep learning has proven to be an extremely powerful paradigm capable of outperforming "traditional" or "conventional" machine learning approaches.
+The recent success behind applying deep learning is partly owed to the fact that DL algorithms are less dependent on manual feature extraction and are capable of achieving high predictive performance when operating on raw-data, such as images and text directly.
```
evancofer: While this is a review/comment piece, all claims must be presented clearly, and supported by concrete information and references to existing literature. This paragraph is a little "handwavy", and doesn't really adhere to that guidance when defining DL and setting it apart from conventional ML.

content/03.ml-concepts.md
evancofer: I disagree, since "careful evaluation" is only a small subset of what we want to drive home here. If the reader remembers one thing from this tip, it should be that "DL is a subfield of ML".

content/03.ml-concepts.md
```diff
@@ -1,24 +1,29 @@
...
+In contrast, traditional machine learning approaches often require manual feature extraction to produce acceptable results. 
+However, the ability of deep learning to perform implicit feature extraction can be regarded as a double-edged sword.
+For instance, those developing deep learning models not only need to ensure that the data contains information relevant to the problem at hand but is also free of systematic biases.
```
evancofer: It seems like this sentence is suggesting that you don't have to make sure your data is "free of systematic biases" if you're not using DL.

content/03.ml-concepts.md
```diff
@@ -1,24 +1,29 @@
...
+As deep learning models are capable of extracting and learning from complex patterns in a training dataset, DL models are also very susceptible to systematic noise and non-salient information, which can not only hamper performance but also lead to spurious conclusions, making proper model evaluation an ever so important topic.
+
+While deep learning models require careful evaluation, good news is that since deep learning is still a subfield of machine learning, the majority of best practices that were developed for machine learning over the last decades also apply to DL.
+Like all computational methods, experiments involving deep learning should be conducted in a systematic manner and accompanied by reproducible instructions as well as rigorous model evaluation.
```
evancofer: The original sentence was not talking about "tests" in the sense of "test sets", but in the sense of "unit tests". Essentially, research code for DL should be testable and free of errors.

It is unclear what "accompanied by reproducible instructions" means. To an outsider, it might sound like a methods section. It might be better to instead mention an entirely reproducible workflow that can be trivially executed.

rasbt: > It is unclear what "accompanied by reproducible instructions" means. 

What I meant was more like instructions to reproduce the results.

Regarding 
> Essentially, research code for DL should be testable and free of errors.

isn't it better to focus on model evaluation instead of re-citing things that apply to research and computing in general?

rasbt:
> I disagree, since "careful evaluation" is only a small subset of what we want to drive home here. If the reader remembers one thing from this tip, it should be that "DL is a subfield of ML".

I see. I am happy to close this PR then. Personally, I didn't like this focus on "DL is a subfield of ML" much as it is obvious and can be stated in one sentence. Initially, I found it was just repeating this point too often and didn't provide anything specific about DL model evaluation, rather, "apply ML best practices" (which some readers may not be familiar with, anyway). 

But that's just my personal impression, and I can see the point behind your arguments. Happy to close this PR as I don't think I can change it in an objective way :)

content/03.ml-concepts.md
rasbt: Yeah, it could sound that way ... the point was more like to provide some extra motivation why this is even more important for DL than for ML, as there is less manual feature engineering involved, which kind of reduces bias (e..g, if I extract facial keypoints and train a classifier, this classifier will obviously not sensitive to skin color or background noise -- the facial keypoint extractor might, but that's a different kind of beast and way too general for an article that aims to discuss deep learning, not computational research in general, in my opinion; i.e., I assume that most people worked with datasets before and read this article primarily because they are interested in DL).

evancofer: Doesn't that make more sense in one of the sections devoted entirely to overfitting or baselines?

evancofer:
@rasbt I think the point of having this section is to make sure that the reader isn't just jumping headfirst into DL without thinking of the bigger picture. The fact that "DL is a subfield of ML" must be established somewhere the article, because much of this general ML knowledge will be critical to the success of would be DL practitioners. By devoting this early tip to that, we clear up any ambiguity and allow the rest of the article to delve deeply into DL-specific advice.  We can't decide if our readers will have experience with general ML or know those best practices, so it's important to point them in the right direction. My personal experience has been that DL is so hyped up that everyone wants to try it, regardless of whether they understand the general ML best practices that they will inevitably need.

I think a good way to make the article more DL-specific would be to also flesh out some of the other tips as well, which focus entirely on DL but may be shorter in length. Some of the discussion of feature extraction might make sense going in the introduction as well. I would love to hear others thoughts on this too (:

agitter:
Note that another opinion on the original form of this tip was left in PR #173.

rasbt:
> The fact that "DL is a subfield of ML" must be established somewhere the article

I agree, but I am still not convinced we need a whole tip on that :).

> We can't decide if our readers will have experience with general ML or know those best practices, so it's important to point them in the right direction. My personal experience has been that DL is so hyped up that everyone wants to try it, regardless of whether they understand the general ML best practices that they will inevitably need.

Not sure how readers will think about it, though. It could easily come across as "YOU shouldn't be using DL, go grab some ML books and come back next year." While I think that the relationship should be mentioned (maybe in the intro), I think that it would be more helpful to summarize/explain the most important things a reader needs to keep in mind (regarding model evaluation approaches) than sending the reader down a rabbit hole. 

> . I think a good way to make the article more DL-specific would be to also flesh out some of the other tips as well, which focus entirely on DL but may be shorter in length. 

Totally agree. My plan was to read the article sequentially and then add things here and there. For some reason, I got distracted and started editing while reading and then ended up making excessive edits here. I agree with some comment above that parts of this tip now better fit into the tip on overfitting.

I'd say the best solution is to roll back this version to the original, and I will try to add some sections of this currently edited version to the overfitting one, because they may fit there more nicely, indeed.

Benjamin-Lee:
Some (disorganized) thoughts, after having been away for a week on break:

>  It could easily come across as "YOU shouldn't be using DL, go grab some ML books and come back next year."

As someone who started out directly using DL with no ML experience (using VAEs), I can say that I really should not have been using DL and should have grabbed some ML books. It wouldn't have been advice I wanted to hear, but it would have saved me a lot of stress and confusion. That being said, there are some types of problems for which DL is the most sensible approach (such as image classification), for which extensive study of "classical" ML algorithms isn't needed. The key thing is understanding generally applicable topics such as overfitting, etc.

I do like @rasbt's proposed name change to "Adhere to machine learning best practices for rigorous model evaluation", as this captures the main message that I think this tip is trying to convey, but I think that it is also alright the way it is currently (albeit less specific).

In the mean time, I am closing this PR since @rasbt says he's alright with closing it and @evancofer thinks the changes would weaken this tip (with which I agree).

rasbt:
> That being said, there are some types of problems for which DL is the most sensible approach (such as image classification), for which extensive study of "classical" ML algorithms isn't needed

I would actually go so far and say that everything that isn't Image, audio, or raw text isn't suited for ML. I.e., everything that is tabular data. Here, you are usually better off with using classic ML approaches. (Personally, I count ~2-3 layer multi-layer perceptrons as classic ML, and deeper MLPs usually don't work anyway in practice). 


 > I can say that I really should not have been using DL and should have grabbed some ML books

From a personal experience, over the last few months, I talked to a lot of people who were asking about advice regarding applying deep learning to a problem (which was often a 1000-10000 record dataset with 5-100 hand-engineered features). When my best and honest advice was to use an SVM, random forest etc., you could literally see their disappointment. In my opinion, there seems to be a huge urge to use DL for the sake of using DL these days. 

Personally, I am still skeptical with regard to devoting a whole tip to "DL is a subfield of ML" -- instead, I would still recommend focusing on problems that DL is good at/problems where DL is a better choice compared to ML and how they are different (hand-engineered features vs automatic feature extraction) and discuss the additional challenges that come up with DL (which is partly covered in the overfitting chapter, but also due to the lack of hand-engineering features, which makes it more susceptible to picking up bias without the user being aware of, which requires particularly careful analyses as it was discussed in this edit).

PR #169, opened 2019-03-14
Tip 2 edits

rasbt:

**Any more details?**

Some edits for tip 2


content/04.baselines.md
```diff
@@ -1,14 +1,16 @@
...
-The performance comparison among DL models and many other ML approaches is informative only when the models are similarly tuned.
+It is worth noting that conventional off-the-shelf machine learning algorithms (e.g., support vector machines and random forests) are also likely to benefit from parameter tuning.
+It can be tempting to train baseline models with these conventional methods using default parameters, which may provide acceptable but not stellar performance, but then tune the settings for DL algorithms to further optimize performance.
+Hu and Greene [@doi:10.1101/385534] discuss a "Continental Breakfast Included" effect by which unequal hyperparameter tuning for different learning algorithms skews the evaluation of these methods, especially when the performance of an algorithm varies substantially with modest changes to its hyperparameters.
+Those wishing to compare of different learning algorithms should tune the parameters of traditional and DL-based methods for predictive performance optimization before making claims about relative performance differences.
```
evancofer: The phrasing here needs some work. The first part should instead be `Those wishing to compare different learning algorithms`. I also think the phrase `tune the parameters of traditional and DL-based methods for predictive performance optimization before` doesn't make any sense. Perhaps just changing it to `tune the parameters of both traditional and DL-based methods to optimize performance before`.

content/04.baselines.md
```diff
@@ -1,14 +1,16 @@
...
-Hu and Greene [@doi:10.1101/385534] discuss a "Continental Breakfast Included" effect by which unequal hyperparameter tuning skews the evaluation of methods, especially those with performance that varies substantially with modest changes to hyperparameters.
-Those wishing to compare methods should tune the parameters of traditional and DL to optimize performance before making claims about relative performance differences.
-The performance comparison among DL models and many other ML approaches is informative only when the models are similarly tuned.
+It is worth noting that conventional off-the-shelf machine learning algorithms (e.g., support vector machines and random forests) are also likely to benefit from parameter tuning.
+It can be tempting to train baseline models with these conventional methods using default parameters, which may provide acceptable but not stellar performance, but then tune the settings for DL algorithms to further optimize performance.
```
evancofer: I would change this back to model, and settings back to parameters. The two are nearly interchangeable in terms of denotation, but I think it's worth describing "algorithmic settings" as parameters since they are specifically chosen (albeit often by a human) based on the training data.

content/04.baselines.md
```diff
@@ -1,14 +1,16 @@
...
-This approach will help researchers with assessing the complexity of the task at hand and debugging more complex DL architectures.
-The utility of these methods is evidenced by the recent development of hybrid models which combine DL and simpler models to improve robustness, interpretability, and confidence estimation [@arxiv:1803.04765; @arxiv:1805.11783].
-Depending on the amount of available data and the type of tasks, DL models may not necessarily perform the best.
-As an illustration, the simple baseline models by Rajkomar et al. [@doi:10.1038/s41746-018-0029-1] achieved performance comparable with that of DL in a number of clinical prediction tasks using electronic health records, which may be a surprise to many.
+Since deep learning requires practitioners to consider a larger number and variety of tuning parameters compared to more traditional methods, it is easy to fall into the trap of performing an unnecessarily convoluted analysis. 
+Hence, before applying deep learning to a given problem, we highly recommend implementing a simple model at the beginning of each study to establish adequate performance baselines.
```
evancofer: I totally agree with this point. It might be better worded a bit differently, since implementation is not always necessary (e.g. sometimes there is an existing baseline to compare to), but there is value in familiarizing ones self with a software framework before applying it to the problem at hand. It may also be worth mentioning encouraging some exploratory data analysis before they even implement/identify a baseline model.

content/04.baselines.md
```diff
@@ -1,14 +1,16 @@
...
+The utility of this approach is evidenced by the recent development of hybrid models that combine DL and simpler models to improve generalization performance, model interpretability, and confidence estimation [@arxiv:1803.04765; @arxiv:1805.11783].
+Depending on the amount and the nature of the available data, as well as the task to be performed, deep learning may not always be able to outperform conventional methods.
+As an illustration, Rajkomar et al. [@doi:10.1038/s41746-018-0029-1] found that simpler baseline models achieved performance comparable with that of DL in a number of clinical prediction tasks using electronic health records, which may be a surprise to many. 
+Another example is provided by Koutsoukas et al., who benchmarked several traditional machine learning approaches against deep neural networks for modeling bioactivity data and found that, while well tuned deep learning approaches tend to outperform conventional classifiers on a moderately sized dataset, simple methods such as Naive Bayes classification tend to outperform deep learning as the noise in the dataset increases [@doi:10.1186/s13321-017-0226-y].
```
evancofer: I haven't read the cited article, but is this supposed to suggest that the `increased noise` in the second case supposed to contrast the `moderate size` of the dataset that DL performs well on? The two aren't mutually exclusive, but the current wording suggests that they are.

content/04.baselines.md
rasbt: > likely to benefit from parameter tuning.

I would rather say change this to "likely to benefit from hyperparameter tuning."

and 

> It can be tempting to train baseline models with these conventional methods using default parameters, 

"to It can be tempting to train baseline models with these conventional methods using default settings," 

since "tuning the parameters of a model" can really mean both, optimizing the weight parameters or changing the learning algorithm's setting

content/04.baselines.md
rasbt: The main point I had in mind was that given a moderately sized dataset (let's assume fixed size), if the dataset is noisy, naive Bayes performed better than DL. Probably not the case as the dataset grows, but that wasn't investigated in the study. I think that rephrasing it to 

> Another example is provided by Koutsoukas et al., who benchmarked several traditional machine learning approaches against deep neural networks for modeling bioactivity data on moderately sized datasets. The researchers found that while well tuned deep learning approaches generally tend to outperform conventional classifiers, simple methods such as Naive Bayes classification tend to outperform deep learning as the noise in the dataset increases [@doi:10.1186/s13321-017-0226-y].

what do you think?

content/04.baselines.md
rasbt: Hm yeah, but I think we do not necessarily want to rely on baselines from literature alone. Imho, we should add your point though, saying that also baselines available from literature should be considered. Reasons for implementing the simple method here are generally

- not all literature results are trustrworthy ;)
- baselines may not be available
- implementations can be complicated
- if a complicated neural net is not training, it may not be clear whether it's due to the settings, implementation errors, or simply weird things happening with the dataset (issues with data prep, etc.)

content/04.baselines.md
evancofer: We should probably  use whatever wording we use throughout the rest of the text. @Benjamin-Lee Once the core content is added, are we planning to edit things for consistency?

content/04.baselines.md
evancofer: Yeah, this change would be perfect. You will need to split into two lines and probably want to move the citation to the first sentence.

content/04.baselines.md
Benjamin-Lee: @evancofer, yes. Once we are happy with the content, I am planning on doing some passes with an eye towards consistency of style, grammar, etc.

content/04.baselines.md
rasbt: oh yes! not sure what happened there -- that sounded terrible!

content/04.baselines.md
rasbt: clarified it in the first sentence now

> ... a larger number and variety of tuning parameters or algorithm settings (so-called hyperparameters) ...

I think it's useful to be clear by what we mean with tuning parameters because the main audience may not be familiar with jargon like hyperparameters.

rasbt:
this should now also be ready for another look

PR #171, opened 2019-03-22
Tip6 edits

rasbt:

**Any more details?**

Made some edits to tip 6. Please help with the discussion regarding the balance of the reconstruction and the distribution loss in the VAE section, which does not seem quite right in the first draft.


rasbt:
(ignore the force bush, messed up and branched from a wrong branched instead of master, and had to rebase, but should be fine now)

PR #125, opened 2019-01-27
first draft for tip 8

tbrittoborges:
**Did you add yourself as a [contributor](https://github.com/Benjamin-Lee/deep-rules/blob/master/contributors.md) if this is your first contribution?**

- [x] Yes, I added myself or am already a contributor

**Any more details?**
This is just a initial draft, I need to check for more recent publications to add to this tip.

PR #172, opened 2019-03-22
Address minor errors from #125 and #170

agitter:

**Any more details?**
(see below)

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -1,6 +1,6 @@
 # Continuous integration / analysis directory
 
-[![Build Status](https://travis-ci.com/manubot/rootstock.svg?branch=master)](https://travis-ci.com/manubot/rootstock)
+[![Build Status](https://travis-ci.org/Benjamin-Lee/deep-rules.svg?branch=master)](https://travis-ci.org/Benjamin-Lee/deep-rules)
 
 This repository uses [continuous analysis](https://doi.org/10.1101/056473 "Reproducible Computational Workflows with Continuous Analysis") to create the manuscript and commit it back to GitHub.
 [`deploy.sh`](deploy.sh) runs on successful `master` branch builds that are not pull requests.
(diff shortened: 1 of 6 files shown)


agitter:
Previous builds had the error `ERROR: No anchor #blackbox for internal URI reference` (e.g. https://travis-ci.org/Benjamin-Lee/deep-rules/builds/509764376#L2191).  Adding a newline after the sub-section header seemed to fix this.  @dhimmel is that the expected behavior?

I also switched the CI badge, which was updated in #170 to the new rootstock CI but not the CI for this repository.

Interestingly, it looks like draft pull requests on GitHub don't trigger CI builds on travis-ci.org.  It works with the new travis-ci.com GitHub app.

agitter:
@dhimmel I'm also wondering whether we should revert the `ci/deploy.sh` changes from `travis-ci.org` to `travis-ci.com`.

dhimmel:
Ah yes these two lines are wrong:

https://github.com/Benjamin-Lee/deep-rules/blob/1b4308438c8675e2480a6cb3dda51d85799806a1/ci/deploy.sh#L52-L53

They should be `.org`... my bad... have made this mistake in many places.

dhimmel:
> Interestingly, it looks like draft pull requests on GitHub don't trigger CI builds on travis-ci.org. It works with the new travis-ci.com GitHub app.

Drafts do trigger CI on travis-ci.com? IIRC draft PRs haven't triggered any travis builds for me. A real annoying bug. I didn't see an issue [here](https://github.com/travis-ci/travis-ci/issues), but we probably should open one.

dhimmel:
> Adding a newline after the sub-section header seemed to fix this. @dhimmel is that the expected behavior?

All the pandoc examples of header identifiers [in the docs](https://pandoc.org/MANUAL.html#header-identifiers) aren't directly followed by text. It doesn't seem to mention a newline is necessary, but perhaps this is intended. It could also be a pandoc bug.

agitter:
> They should be .org... my bad... have made this mistake in many places.

No problem, updated in fae4239

> Drafts do trigger CI on travis-ci.com?

They triggered on this WIP draft pull request https://github.com/gitter-lab/SCINGE/pull/13  I'm not sure what we did differently there to get it to work.

PR #192, opened 2019-06-10
indicate accountability

ttriche:
Will need to edit final draft for this to be strictly true, which is the intent; recent remark about LR vs. DL (or, for that matter, RF/similar ML) is worth emphasizing.  If the training data is insufficiently informative, or lacks structure, the marginal benefit of DL relative to the marginal effort involved training DNN/CNN type models is usually difficult to justify.

- [ x ] Yes, I added myself or am already a contributor

**Any more details?**
Per issue, I have indicated my accountability for what I contribute to the paper.


Benjamin-Lee:
Don't forget to add your information to metadata.yaml!

PR #213, opened 2020-06-05
Close #210

### LINKED ISSUE #210: Should we change the overfitting figure?
One thing that I've been thinking about is modifying the overfitting figure, first discussed in #113. Right now, it looks like this:

One thing that is bothering me is that the testing data is different (in a separate range) from the training data. Would it make more sense to have them be in the same range? That way, the overfit model wouldn't be anywhere near the points. We might need to reduce the number of points to ensure that the point (pun intended) of the graph is clear since the plotted points are quite large for readability.

Also, should we remove the xkcd style plotting? I used it at the suggestion of _PLOS Comp Bio_ editor Philip Bourne who [wrote](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1003833):

> Using a sketch style conveys to the viewer that the data is approximate, and that it is the higher-level concepts rather than low-level details that are important

and wrote in [Ten Simple Rules for Writing a PLOS Ten Simple Rules Article](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1003858):

> There are few scientific articles where you can borrow your plot style from the XKCD comic series (http://xkcd.com/), and this article is one of them.

Benjamin-Lee:

**Any more details?**

@rasbt Can you take a look? I was just rereading your Python Machine Learning book and really liked your figure on overfitting, so what are your thoughts on #210 and this figure?


agitter:
From #210 
> One thing that is bothering me is that the testing data is different (in a separate range) from the training data. Would it make more sense to have them be in the same range?

I thought this was an important point.  Having the testing points from a different range is more like illustrating a dataset shift, or training and testing points that are drawn from different distributions.  I actually think that is also a very important concept to convey.  But it may be better illustrated in a second conceptual figure like this so that one figure can highlight overfitting and another can highlight dataset shift.

I do like the stylistic changes here, and the xkcd style plotting in general.

rasbt:
You mean this one?

Screen Shot 2020-06-05 at 8 42 41 AM

It looks kind of old to me, but well, I have probably seen it too many times in the last 6 years :P. Regarding the discussion in the other thread (#210), we may have to figure out whether we want to convey overfitting & out-of-training-range issues in the same figure, or whether we want to have two separate illustrations. Let's resolve this discussion in the other thread first and then revisit this one. 

Spontaneously, I would suggest using the other figure as illustration of the out-of-range issue, and this one (modified with XKCD style) can be used as illustration of just overfitting.

Benjamin-Lee:
Just pushed a new version that only shows lack of generalization to out of training set data, which fits with the example in the text of the chest x-rays from different hospitals breaking the DL model.

Benjamin-Lee:
image

It also features fewer points than before which prevents the peaks of the polynomial from going out of the y range.

pstew (approved):

PR #216, opened 2020-06-09
Update metadata.yaml

tbrittoborges:
Update metadata.yaml with tbrittoborges affiliation

**Any more details?**


content/metadata.yaml
```diff
@@ -130,4 +130,12 @@ authors:
     affiliations:
       - Center for Epigenetics, Van Andel Research Institute
       - Department of Translational Genomics, Keck School of Medicine, University of Southern California
+   -
```
agitter, line 133:
```suggestion
  -
```
This may be enough to fix the build error

Issue #208, opened 2020-06-04
Fix VAE hyperparameter component importance TODO


Benjamin-Lee:
@rasbt, you raised this this comment (see [here](https://github.com/Benjamin-Lee/deep-rules/blob/9522ea5b6db67fd06f3a98ffd4c7ce518e7967aa/content/08.hyperparameters.md#L12) for context):

 > I am not sure this is correct, why would a larger number of, e.g., pixels make the reconstruction loss more important? I suppose this is true if we just sum over the pixel-wise differences, but if we average, e.g., using MSE, I am not convinced this is true. Please comment.

Would you mind taking the lead on resolving this? I'm not sure I fully understand either the sentence in question (which should be clarified regardless of whether it is correct) nor the question itself.

cgreene:
I agree that it is _possible_ to do it with averaging, but I would argue that it is rarely done in practice. Averaging would add an additional step that could slow the computation.

Also, here are a number of Keras and pytorch examples, in which this is not done and these are among the first hits.

<img width="519" alt="Screen Shot 2020-06-04 at 7 37 54 AM" src="https://user-images.githubusercontent.com/542643/83752495-a2eb5d00-a636-11ea-93ea-dd8b1a7a4e5a.png">

<img width="750" alt="Screen Shot 2020-06-04 at 7 38 04 AM" src="https://user-images.githubusercontent.com/542643/83752526-ad0d5b80-a636-11ea-8465-6b49fe601517.png">

<img width="630" alt="Screen Shot 2020-06-04 at 7 38 42 AM" src="https://user-images.githubusercontent.com/542643/83752536-b1397900-a636-11ea-847c-60cd5b210f29.png">

<img width="788" alt="Screen Shot 2020-06-04 at 7 38 55 AM" src="https://user-images.githubusercontent.com/542643/83752550-b4cd0000-a636-11ea-9025-3205085840b6.png">

While we might wish that something would be trivial and desirable, it doesn't mean that it is the case :)

cgreene:
https://pytorch.org/docs/master/nn.functional.html#binary-cross-entropy

```
reduction (string, optional) – Specifies the reduction to apply to the output: 'none' | 'mean' | 'sum'. 'none': no reduction will be applied, 'mean': the sum of the output will be divided by the number of elements in the output, 'sum': the output will be summed. Note: size_average and reduce are in the process of being deprecated, and in the meantime, specifying either of those two args will override reduction. Default: 'mean'
```
[...]

PR #217, opened 2020-09-16
Remove mention of VAE architecture as a hyperparameter

Benjamin-Lee:
I'm working on getting the manuscript ready for submission. We've resolved almost all of the issues within the text except for this one. I propose removing the section altogether. Although architecture tuning is a good idea, I'm not sure that the VAE example is right. I think we're best off to remove it.


Benjamin-Lee:
This is in reference to #208, by the way.

rasbt:
I agree, I think VAE are becoming less relevant for practical applications (although there are some interesting new advances) and are not widely used in comp-bio. I agree that removing it makes sense. But maybe someone else can also share his or her thoughts on this before making a definite decision here.

rasbt (approved):

cgreene (approved):
Strong agree 👍

PR #219, opened 2020-09-16
Close #202

### LINKED ISSUE #202: Biological sequence data is related in complex ways and this makes validation difficult
This is a cool project and the draft is looking nice.

I have one thing I would add, probably to "Tip 7: Address deep neural networks' increased tendency to overfit the dataset". I'm mentioning it here for discussion, and if wanted I would be happy to add some text describing it.

When splitting a dataset of biological sequences or structures, care should be taken that there is no evolutionary relationship between sequences in the training set, sequences in the validation set and sequences in the test set. Many people split proteins into datasets using a threshold of 30% sequence identity, i.e. the training and validation sets will not share any sequence that is 30% or more similar. However, it is known that many proteins share homology down to effectively 0% sequence identity - see Figure 1 of [Chothia and Lesk 1986](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC1166865/) for example.

Poor dataset splitting means that the method being benchmarked appears to have better performance than it does, as one is partially measuring an ability to detect homologs. This problem affects protein secondary structure prediction, tertiary contact prediction, protein design studies, in fact almost anywhere protein data is used for machine learning. One way round it is to use databases such as CATH and ECOD to split sequences based on structural and evolutionary relationships.

Tagging @Benjamin-Lee as I think he drafted this section.

Benjamin-Lee:
This PR closes #202 by specifically mentioning the dangers of splitting sequence data.

### DIFF
--- a/content/09.overfitting.md
+++ b/content/09.overfitting.md
@@ -17,7 +17,7 @@ Another way, as described by Chuang and Keiser, is to identify the baseline leve
 If the model performs no better on real data than randomly scrambled data, then the performance of the model can be attributed to overfitting.
 
 Additionally, in biology and medicine it is critical to consider independence when defining training and test sets.
-For example, a DL model for pneumonia detection in chest X-rays performed well but failed to generalize to outside hospitals because they were able to detect which hospital the image was from and adjust accordingly [@doi:10.1371/journal.pmed.1002683].
+For example, a DL model for pneumonia detection in chest X-rays performed well but failed to generalize to outside hospitals because they were able to detect which hospital the image was from and exploited this information when making predictions [@doi:10.1371/journal.pmed.1002683].
 Similarly, when dealing with sequence data, holding out data that are evolutionarily related or share structural homology to the training data can result in overfitting.
 In these cases, simply holding out test data selected from a random partition of the training data is insufficient.
 The best remedy for confounding variables is to [know your data](#know-your-problem) and to test your model on truly independent data.
--- a/content/09.overfitting.md
+++ b/content/09.overfitting.md
@@ -12,7 +12,7 @@ The simplest way to combat overfitting is to detect it.
 This can be done by splitting the dataset into three parts: a training set, a tuning set (also commonly called a validation set in the machine learning literature), and a test set.
 By exposing the model solely to the training data during fitting, a researcher can use the model's performance on the unseen test data to measure the amount of overfitting.
 While a slight drop in performance from the training set to the test set is normal, a significant drop is a clear sign of overfitting (see Figure @fig:overfitting-fig for a visual demonstration of an overfit model that performs poorly on test data).
-Additionally, there are a variety of techniques to reduce overfitting during training including data augmentation and regularization techniques such as dropout [@url:http://jmlr.csail.mit.edu/papers/v15/srivastava14a.html] and weight decay [@tag:krogh-weight-decay].
+In addition, there are a variety of techniques to reduce overfitting during training including data augmentation and regularization techniques such as dropout [@url:http://jmlr.csail.mit.edu/papers/v15/srivastava14a.html] and weight decay [@tag:krogh-weight-decay].
 Another way, as described by Chuang and Keiser, is to identify the baseline level of memorization of the network by training on the data with the labels randomly shuffled and to see if the model performs better on the actual data [@doi:10.1021/acschembio.8b00881].
 If the model performs no better on real data than randomly scrambled data, then the performance of the model can be attributed to overfitting.
 
@@ -23,4 +23,4 @@ In these cases, simply holding out test data selected from a random partition of
 The best remedy for confounding variables is to [know your data](#know-your-problem) and to test your model on truly independent data.
 
 In essence, practitioners should split data into training, tuning, and single-use testing sets to assess the performance of the model on data that fully assesses generalization (_e.g._ by holding out all data from one hospital as a test set).
-Additionally, be cognizant of the danger of skewed or biased data artificially inflating accuracy.
+Futhermore, be cognizant of the danger of skewed or biased data artificially inflating accuracy.
--- a/content/09.overfitting.md
+++ b/content/09.overfitting.md
@@ -22,5 +22,5 @@ Similarly, when dealing with sequence data, holding out data that are evolutiona
 In these cases, simply holding out test data selected from a random partition of the training data is insufficient.
 The best remedy for confounding variables is to [know your data](#know-your-problem) and to test your model on truly independent data.
 
-In essence, practitioners should split data into training, tuning, and single-use testing sets to assess the performance of the model on data that fully assesses generalization (_e.g._ by holding out all data from one hospital as a test set).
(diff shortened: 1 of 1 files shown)


content/09.overfitting.md
```diff
@@ -16,9 +16,11 @@ Additionally, there are a variety of techniques to reduce overfitting during tra
 Another way, as described by Chuang and Keiser, is to identify the baseline level of memorization of the network by training on the data with the labels randomly shuffled and to see if the model performs better on the actual data [@doi:10.1021/acschembio.8b00881].
 If the model performs no better on real data than randomly scrambled data, then the performance of the model can be attributed to overfitting.
 
-Additionally, one must be sure that their dataset is not skewed or biased, such as by having confounding and scientifically irrelevant variables that the model can pick up on [@doi:10.1371/journal.pmed.1002683].
-In this case, simply holding out test data is insufficient.
+Additionally, one must be sure that their dataset is not skewed or biased, such as by having confounding and scientifically irrelevant variables that the model can pick up on.
```
cgreene, line 19: Do we want to say _independent_ and identically distributed? It may be helpful to include the slightly more technical terminology here as it would connect to the broader literature on the topic.

cgreene, line 19:
```suggestion
Additionally, in biology and medicine it is critical to consider independence when defining training and test sets.
```

content/09.overfitting.md
```diff
@@ -16,9 +16,11 @@ Additionally, there are a variety of techniques to reduce overfitting during tra
...
+For example, a DL model for pneumonia detection in chest X-rays performed well but failed to generalize to outside hospitals because they were able to detect which hospital the image was from and adjust accordingly [@doi:10.1371/journal.pmed.1002683].
+Similarly, when dealing with sequence data, holding out data that are evolutionarily related or share structural homology to the training data can result in overfitting.
+In these cases, simply holding out test data is insufficient.
```
cgreene, line 22:
```suggestion
In these cases, simply holding out test data selected from a random partition of the training data is insufficient.
```

content/09.overfitting.md
```diff
@@ -16,9 +16,11 @@ Additionally, there are a variety of techniques to reduce overfitting during tra
...
 The best remedy for confounding variables is to [know your data](#know-your-problem) and to test your model on truly independent data.
 
-In essence, split your data into training, tuning, and single-use testing sets to assess the performance of the model on data it truly has not seen before. 
+In essence, split your data into training, tuning, and single-use testing sets to assess the performance of the model on data it truly has not seen before.
```
cgreene, line 25:
```suggestion
In essence, practitioners should split data into training, tuning, and single-use testing sets to assess the performance of the model on data that fully assesses generalization (i.e., by holding out all data from one hospital as a test set).
```

cgreene (approved):
I tried to make this a little more concrete. See what you think. I also think we should use the word independent when we get into this discussion.

content/09.overfitting.md
Benjamin-Lee:
```suggestion
In essence, practitioners should split data into training, tuning, and single-use testing sets to assess the performance of the model on data that fully assesses generalization (_e.g._ by holding out all data from one hospital as a test set).
```

content/09.overfitting.md
Benjamin-Lee: For ensuring generalization, is identical distribution a hard requirement? In the hospital example, independent hospitals had different rates of pneumonia which should not have affected the model if it were truly learning the signature of pneumonia rather than the relative frequencies.

content/09.overfitting.md
Benjamin-Lee: Two questions:

1. Should this be an e.g. rather than an i.e.?
2. Does it make sense to specifically refer to an example in the closing line? We haven't done so before but I don't see a strong reason not to.

content/09.overfitting.md
cgreene: I definitely think independent is the bigger issue here. Identically distributed is helpful for performance but if your test set is not identically distributed in the same way that other generalization sets are not, I guess you should still get an accurate performance estimate.

Benjamin-Lee: Ok this is fixed as:

> Additionally, in biology and medicine it is critical to consider independence when defining training and test sets.

Benjamin-Lee:
@cgreene thanks for the suggestion on independence. It's merged into this PR now.

content/09.overfitting.md
```diff
@@ -12,13 +12,15 @@ The simplest way to combat overfitting is to detect it.
...
 Another way, as described by Chuang and Keiser, is to identify the baseline level of memorization of the network by training on the data with the labels randomly shuffled and to see if the model performs better on the actual data [@doi:10.1021/acschembio.8b00881].
 If the model performs no better on real data than randomly scrambled data, then the performance of the model can be attributed to overfitting.
 
-Additionally, one must be sure that their dataset is not skewed or biased, such as by having confounding and scientifically irrelevant variables that the model can pick up on [@doi:10.1371/journal.pmed.1002683].
-In this case, simply holding out test data is insufficient.
+Additionally, in biology and medicine it is critical to consider independence when defining training and test sets.
+For example, a DL model for pneumonia detection in chest X-rays performed well but failed to generalize to outside hospitals because they were able to detect which hospital the image was from and adjust accordingly [@doi:10.1371/journal.pmed.1002683].
```
rasbt, line 20: >  and adjust accordingly 

I am not quite sure what the "adjust accordingly" means in this context. Does it mean that the researchers adjusted the model, or did the model adjust something (I presume the prediction)? 

> For example, a DL model for pneumonia detection in chest X-rays performed well but failed to generalize to images collected at outside hospitals because they were able to detect which hospital the image was from **and exploited this information when making predictions.**

content/09.overfitting.md
```diff
@@ -12,13 +12,15 @@ The simplest way to combat overfitting is to detect it.
...
+Similarly, when dealing with sequence data, holding out data that are evolutionarily related or share structural homology to the training data can result in overfitting.
+In these cases, simply holding out test data selected from a random partition of the training data is insufficient.
 The best remedy for confounding variables is to [know your data](#know-your-problem) and to test your model on truly independent data.
 
-In essence, split your data into training, tuning, and single-use testing sets to assess the performance of the model on data it truly has not seen before. 
-Additionally, be cognizant of the danger of skewed or biased data artificially inflating accuracy.
+In essence, practitioners should split data into training, tuning, and single-use testing sets to assess the performance of the model on data that fully assesses generalization (_e.g._ by holding out all data from one hospital as a test set).
```
rasbt, line 25: > that fully assesses generalization

The (statistical) definition of generalization performance is based on the whole population, and we are still only approximating or estimating it. I suggest a slight rewording

> In essence, practitioners should split data into training, tuning, and single-use testing sets to assess the performance of the model on data that can provide a reliable estimate of its generalization performance

rasbt (commented):
Just a few minor points

Benjamin-Lee:
@rasbt, thanks for the suggestions. Merged into the PR.

rasbt (approved):

PR #220, opened 2020-09-16
Update 06.know-your-problem.md

cgreene:
**Did you add yourself as a [contributor](https://github.com/Benjamin-Lee/deep-rules/blob/master/contributors.md) if this is your first contribution?**

- [x] Yes, I added myself or am already a contributor

**Any more details?**

I am not always a big fan of data simulation, but it probably merits a mention. It was in tip 1 (removed in #218) which appears to be the wrong place. This might be a better spot.

PR #229, opened 2020-09-29
editing to point 1

ejfertig:

- [ x] Yes, I added myself or am already a contributor

**Any more details?**

### DIFF
--- a/content/03.ml-concepts.md
+++ b/content/03.ml-concepts.md
@@ -5,7 +5,8 @@ DL has proven to be an extremely powerful paradigm capable of outperforming “t
 Many best practices for machine learning also apply to deep learning.
 Like all computational methods, deep learning should be applied in a systematic manner that is reproducible and rigorously tested.
 
-Those developing deep learning models should select datasets to train and test model performance that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions. For example, supervised deep learning for phenotype prediction should be applied to datasets that contain large numbers of representative samples from all phenotypes to be predicted.
+Those developing deep learning models should select datasets to train and test model performance that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions. 
+For example, supervised deep learning for phenotype prediction should be applied to datasets that contain large numbers of representative samples from all phenotypes to be predicted.
 Biases in testing data can also unduly influence measures of model performance, and it may be difficult to directly identify confounders from the model.
 Investigators should consider the extent to which the outcome of interest is likely to be predictable from the input data and begin by thoroughly inspecting the input data.
 Suppose that there are robust heritability estimates for a phenotype that suggest that the genetic contribution is modest but a deep learning model predicts the phenotype with very high accuracy.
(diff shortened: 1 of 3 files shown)


content/06.know-your-problem.md
```diff
@@ -24,7 +24,7 @@ As a consequence, models may learn to rely on correlations that are irrelevant i
 Other study design considerations that should not be overlooked include knowing whether a study involves biological or technical replicates or both.
 For example, are some samples collected from the same individuals at different time points?
 Are those time points before and after some treatment?
-If one assumes that all the samples are independent but that is in fact not the case, a variety of issues may arise, including having a lower effective sample size than expected.
+If one assumes that all the samples are independent but that is in fact not the case, a variety of issues may arise, including having a lower effective sample size than expected. As described in Tip 1, unsupervised learning and other exploratory analyses can be identify such biases in these datasets prior to applying the deep learning model.
```
cgreene, line 27:
```suggestion
If one assumes that all the samples are independent but that is in fact not the case, a variety of issues may arise, including having a lower effective sample size than expected.
As described in Tip 1, unsupervised learning and other exploratory analyses can be identify such biases in these datasets prior to applying the deep learning model.
```

cgreene (approved):
I one sentence per lined this and otherwise like the change 👍

content/03.ml-concepts.md
```diff
@@ -22,6 +25,9 @@ Also, many conventional metrics for classification (e.g. area under the receiver
 Model performance should be evaluated with a carefully picked panel of relevant metrics that make minimal assumptions about the composition of the testing data [@doi:10.1021/acs.molpharmaceut.7b00578], with particular consideration given to metrics that are most directly applicable to the task at hand.
 
 Extreme cases warrant testing the robustness of the model and metrics on simulated data for which the ground truth is known.
-Said simulations can be used to verify the correctness of the model’s implementation as well.
+To accurately test the performance of the model, it is important that simulated datasets be generated for a range of parameters.
+Simulated datasets generated from parameters that are consistent with the assumptions of the deep learning models can be used to verify the correctness of the model’s implementation.
+Further varying the parameters to test variation and violations of model assumptions can further test the sensitivity of performance to those assumptions.
```
Benjamin-Lee, line 30: This sounds a bit repetitive: "Further varying the parameters to test variation"

Maybe:
```suggestion
Further modifying the parameters to test variation and violations of model assumptions can further test the sensitivity of performance to those assumptions. 
```

content/03.ml-concepts.md
```diff
@@ -22,6 +25,9 @@ Also, many conventional metrics for classification (e.g. area under the receiver
...
 
 Extreme cases warrant testing the robustness of the model and metrics on simulated data for which the ground truth is known.
-Said simulations can be used to verify the correctness of the model’s implementation as well.
+To accurately test the performance of the model, it is important that simulated datasets be generated for a range of parameters.
+Simulated datasets generated from parameters that are consistent with the assumptions of the deep learning models can be used to verify the correctness of the model’s implementation.
+Further varying the parameters to test variation and violations of model assumptions can further test the sensitivity of performance to those assumptions. 
+Generative models based upon sample training sets can provide important tools to generate large cohorts of simulated datasets that are representative of the real world problems to which the machine learning prediction models will be applied.
```
Benjamin-Lee, line 31: This is a good paragraph, just not 100% sure it belongs here. For context, in #218, I proposed removing the paragraph altogether. In response, @cgreene created #220, writing

> I am not always a big fan of data simulation, but it probably merits a mention. It was in tip 1 (removed in #218) which appears to be the wrong place. [Tip 4] might be a better spot.

I propose closing #218 and moving this paragraph to Tip 4. Does that work for you @ejfertig and @cgreene?

cgreene: I agree with the proposed change of putting data simulation into tip 4. What do you think about merging it here first and then moving/integrating into tip 4 as a second step of edits?

pstew: Concur @cgreene

cgreene: Ok - I am going to mark approve which will enable @Benjamin-Lee to merge this if desired and then open up the PR for the second set. I'm in meetings all day today but can try to get to it tomorrow if nobody else ✅ s it first.

cgreene (approved):

content/03.ml-concepts.md
cgreene: whoops - I had already marked approve 😂  - i guess someone else has to 😁

pstew (approved):

content/03.ml-concepts.md
Benjamin-Lee: I'll merge this and do the PR for the move to tip 4. Thanks!

Issue #228, opened 2020-09-29
Rename Tip 1 to "Decide if deep learning may be appropriate for your problem"


ejfertig:
I find the comment "Those developing deep learning models should select datasets that are relevant to the problem at hand; non-salient data can hamper performance or lead to spurious conclusions." unclear. How can someone determine if they have the right dataset? Does this mean in terms of sample size or something else? Some clear examples would help.

ejfertig:
Also, as a further comment to this -- there appears to be some overlap between the dataset selection w/in tip 1 and the tip 4 know your data. Would help to streamline and clarify the differences in the points on data selection in these two tips.

cgreene:
This is a good point - particularly as it relates to transfer learning and/or domain adaptation.

Benjamin-Lee:
@cgreene what do you think about moving the content that overlaps in tip 1 to each respective tip (overfitting and know your data, etc) and then subbing in the "decide if DL is right for your problem" tip you proposed here: https://github.com/Benjamin-Lee/deep-rules/pull/225#pullrequestreview-500191626

cgreene:
Yea - I think I really like this. Tip 1 could be "Decide if deep learning may be appropriate for your problem" and then the rest of the text moves. That tip could contain the current sentiment of tip 1. 👍

ejfertig:
Agreed. I think it would be really helpful to define different flavors of deep learning here and how you can select between models for specific problems. How can you use DL in a supervised context? Unsupervised? When would DL be the best choice?

Benjamin-Lee:
@cgreene Do you think you could work together with me to get the rewrite done by the end of the week, which is the original deadline for freezing changes (#226)? We could also push the deadline back, though I hate to prolong things even more. 

Either way, I can make a PR moving the content and renaming the first tip to get started on this process. @pstew, @agitter, and @rasbt, what do you think of this idea?

ejfertig:
I can help with editing and a view fOr clarity from the non DL perspective if it’s helpful.

agitter:
> what do you think of this idea?

I have other deadlines this week so I'm not sure I'll be able to spend much time here.  Please feel free to made edits as you'd like.

Benjamin-Lee:
@agitter, understood. Would pushing back another week be better for your schedule? I'm not trying to rush things, just to keep up the momentum.

agitter:
It may not make a difference.  You have some good momentum now, so I'd say keep it going.

[...]

Issue #15, opened 2018-11-06
Discuss computational issues


SiminaB:
Mention compute time issues, tools that are available, open source programming. The "Ten Quick Tips" for comp bio article linked in #1 specifically has "Program your software with open source code and platforms" as tip 9, so we have to make it somewhat DL-specific (eg how much can be done in R vs needing TensorFlow, how much can actually be done on a PC vs needing access to a cluster.) Also figure out the best way to create an actual usable rule around this.

fmaguire:
Tangentially referred to in https://github.com/Benjamin-Lee/deep-rules/blob/master/content/05.dl-complexities.md 

> Despite improved framework ease-of-use and on-demand cloud computing resources this means DL can be very demanding, often requiring significant infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740].

Do we want to add more?

PR #241, opened 2020-10-07
New Tip 1

Benjamin-Lee:

**Any more details?**

Per the discussion in #228, this PR moves the content in the current tip 1 around (only removing a few lines) and creates space for a new first tip focusing on whether you should use DL in the first place.


cgreene:
I think this is going in a good direction. What are you thinking re: how to decide? For me it comes down to:
* Do we expect that the features are not already heavily optimized for the task.
* Are we ok with the features being combined in somewhat difficult to untangle ways to carry out the task?
* Do we have one or more plausible strategies for dealing with the large number of parameters?
* Does it matter that there are likely an arbitrary number of equivalent solutions?

SiminaB:
I'm reading through the manuscript right now. Sounds like a good plan!
A couple of related comments: 

- Part of tip 2 should also now go into Tip 1, since a simpler ML model may suffice. Eg the second paragraph:

> Depending on the amount and the nature of the available data, as well as the task to be performed,
deep learning may not always be able to outperform conventional methods. As an illustration,
Rajkomar et al. [13] found that simpler baseline models achieved performance comparable with that
of DL in a number of clinical prediction tasks using electronic health records, which may be a surprise
to many. Another example is provided by Koutsoukas et al., who benchmarked several traditional
machine learning approaches against deep neural networks for modeling bioactivity data on
moderately sized datasets [14]. The researchers found that while well tuned deep learning
approaches generally tend to outperform conventional classiers, simple methods such as Naive
Bayes classication tend to outperform deep learning as the noise in the dataset increases.

- I'm not a huge fan of the heritability example, because there's no citation. Like, this could be a problem, sure, but has it ever been a problem in practice? We can use reference 39 as an example instead (or another similar example). This is also related to Tip 8, since if you can't interpret the models you don't even know what possible confounders may have gotten included.

Benjamin-Lee:
Converting this from draft status to be ready for review.

Benjamin-Lee:
BTW, this closed #228

chevrm (changes requested):
New lines 3 & 4 still needs some work.

Benjamin-Lee:
> New lines 3 & 4 still needs some work.

What do you have in mind? I'm trying to get at the misleading conception that since NN are a universal function approximators  they should be used universally.

chevrm:
I think the sentiment works; just needs some tweaking of the language. Maybe something like

In recent years, the number of publications implementing DL in biology have risen tremendously.
Given DL's usefulness across a range of scientific questions and data modalities, it may appear that it is capable of anything.
Neural networks are universal function approximators, meaning that they are in principle capable of learning any function [@doi:10.1007/BF02551274; @tag:hornik-approximation].
If DL is so powerful and popular, why would one ever not choose to use it?

I still don't love the "it may appear that it is capable of anything" bit. Any suggestions?

Benjamin-Lee:
> I still don't love the "it may appear that it is capable of anything" bit. Any suggestions?

What about something like "a panacea for modeling problems"?

chevrm:
> > I still don't love the "it may appear that it is capable of anything" bit. Any suggestions?
> 
> What about something like "a panacea for modeling problems"?

Yes, I think that works well

chevrm (approved):

Benjamin-Lee:
Also note that this new tip discusses resource constraints, as mentioned in #15 and #82

Benjamin-Lee:
In further looking at the issue history to make sure we didn't miss anything, what about renaming this tip "Don't use DL if you don't have to"? In fact, that was one of the earliest tip proposals we had: #3

For the tip's name, we have a few options:
1. Decide whether your problem is appropriate for deep learning (this PR)
2. Decide whether deep learning is appropriate for your problem
3. Don't use deep learning if you don't have to

What are people's thoughts?

content/02.intro.md
```diff
@@ -9,7 +9,11 @@ Since DL is an active and specialized research area, detailed resources are rapi
 To address this issue, we solicited input from a community of researchers with varied biological and deep learning interests to write this manuscript collaboratively using the GitHub version control platform [@url:https://github.com/Benjamin-Lee/deep-rules] and Manubot [@doi:10.1371/journal.pcbi.1007128].
 
 Through the course of our discussions, several themes became clear: the importance of understanding and applying ML fundamentals [@doi:10.1186/s13040-017-0155-3] as a baseline for utilizing DL, the necessity for extensive model comparisons with careful evaluation, and the need for critical thought in interpreting results generated by means of DL, among others.
-Both DL models and the datasets to which they are applied impact prediction results, so it is important to consider both when generating biological or clinical insights from these methods.
+The major similarities between deep learning and traditional computational methods also became apparent.
```
SiminaB, line 12: Should use "DL" and "ML" abbreviations after they are introduced.

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+If DL is so powerful and popular, why would one ever not choose to use it?
+
+The reason is simple: DL is not suited to every situation in reality.
+Training DL models requires a significant amount of data, computing power, and expertise.
+In some areas of biology where data collection is thoroughly automated, such as DNA sequencing, large amounts of quality data may be available.
+For other areas which rely on manual data collection, there may not be enough data to effectively train models.
+Though there are methods to increase the amount of training data such as data augmentation (in which existing data is slightly manipulated to yield "new" samples) and weak supervision (in which simple labeling heuristics are combined to produce noisy, probabilistic labels) [@arxiv:1605.07723v3], these methods cannot overcome a complete shortage of data.
+As a rule of thumb, DL should only be considered for datasets with at least one thousand samples, though it is best suited to cases when datasets contain orders of magnitude more samples.
```
SiminaB, line 13: Should have a reference for this rule of thumb.

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+
+Furthermore, training DL models can be very demanding, often requiring extensive computing infrastructure and patience to achieve state-of-the-art performance [@doi:10.1109/JPROC.2017.2761740].
+In some DL contexts, such as generating human-like text, state-of-the-art models have over one hundred billion parameters [@arxiv:2005.14165].
+Training such large models from scratch can be a costly and time-consuming undertaking [@arxiv:1906.02243].
+Luckily, most DL research in biology will not require nearly as much computation, and there are methods for reducing the amount of training required in some cases (described in [Tip 5](#architecture)).
```
SiminaB, line 18: Perhaps note here that although they will not require as much computation, they still generally require some level of cloud computing.

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+Specialized hardware such as discrete graphics processing units (GPUs) or custom DL accelerators can dramatically reduce the time and cost required to train models, but this hardware is not universally accessible.
+Currently, both GPU- and DL-optimized accelerator-equipped servers can be rented from cloud providers, though working with these servers adds additional cost and complexity.
+As DL becomes more popular, DL-optimized accelerators are likely to be more broadly available (for example, recent-generation iPhones already have such hardware).
+In contrast, traditional ML training can often be done on a laptop (or even a \$5 computer [@arxiv:1809.00238]) in seconds to minutes.
+
+Beyond the necessity for greater data and computational capacity in DL, building and training DL models generally requires more expertise than traditional ML models.
+Currently, there are several competing programming frameworks for DL such as Tensorflow [@arxiv:1603.04467] and PyTorch [@arxiv:1912.01703].
+These frameworks allow users to create and deploy entirely novel model architectures and are widely used in DL research as well as in indutry.
```
SiminaB, line 26: indutry -> industry. Maybe better to say "industrial applications" than "industry" since research can be done in industry as well?

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+Projects such as Keras [@https://keras.io], a high-level interface for TensorFlow, make it relatively straightforward to design and test custom DL architectures.
+In the future, projects such as these are likely to bring DL experimentation within reach to even more researchers.
+
+There are some types of problems in which using DL is strongly indicated over ML.
+Assuming a sufficient quantity of quality data is available, applications such as computer vision and natural language processing are likely to benefit from DL.
+In fact, these areas were the first to see significant breakthroughs through the application of DL [@doi:10.1145/3065386] during the recent "DL revolution."
+For example, Ferreira et al. used DL to recognize individual birds from images [@doi:10.1111/2041-210X.13436].
+This problem was historically difficult but, by combining automatic data collection using RFID tags with data augmentation and transfer learning (explained in [Tip 5](#architecture), the authors were able to use DL achieve 90% accuracy in several species.
```
SiminaB, line 40: I think there should be another parentheses ")" after "(#architecture)" to render correctly.

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+Other areas include generative models, in which new samples are able to be created based on the training data, and reinforcement learning, in which agents are trained to interact with their environments.
+In general, before using DL, investigate whether similar problems (including analogous ones in other domains) have been solved successfully using DL.
+
+Depending on the amount and the nature of the available data, as well as the task to be performed, DL may not always be able to outperform conventional methods.
+As an illustration, Rajkomar et al. [@doi:10.1038/s41746-018-0029-1] found that simpler baseline models achieved performance comparable with that of DL in a number of clinical prediction tasks using electronic health records, which may be a surprise to many.
+Another example is provided by Koutsoukas et al., who benchmarked several traditional machine learning approaches against deep neural networks for modeling bioactivity data on moderately sized datasets [@doi:10.1186/s13321-017-0226-y].
```
SiminaB, line 46: "machine learning" -> "ML"

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+For example, Ferreira et al. used DL to recognize individual birds from images [@doi:10.1111/2041-210X.13436].
+This problem was historically difficult but, by combining automatic data collection using RFID tags with data augmentation and transfer learning (explained in [Tip 5](#architecture), the authors were able to use DL achieve 90% accuracy in several species.
+Other areas include generative models, in which new samples are able to be created based on the training data, and reinforcement learning, in which agents are trained to interact with their environments.
+In general, before using DL, investigate whether similar problems (including analogous ones in other domains) have been solved successfully using DL.
+
+Depending on the amount and the nature of the available data, as well as the task to be performed, DL may not always be able to outperform conventional methods.
+As an illustration, Rajkomar et al. [@doi:10.1038/s41746-018-0029-1] found that simpler baseline models achieved performance comparable with that of DL in a number of clinical prediction tasks using electronic health records, which may be a surprise to many.
```
SiminaB, line 45: Not sure we need "which may be a surprise to many."

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+Another example is provided by Koutsoukas et al., who benchmarked several traditional machine learning approaches against deep neural networks for modeling bioactivity data on moderately sized datasets [@doi:10.1186/s13321-017-0226-y].
+The researchers found that while well tuned DL approaches generally tend to outperform conventional classifiers, simple methods such as Naive Bayes classification tend to outperform DL as the noise in the dataset increases.
+Similarly, Chen et al. [@doi:s41746-019-0122-0] tested DL and a variety of traditional ML methods such as logistic regression and random forests on five different clinical datasets, finding that the non-DL methods matched or exceeded the accuracy of the DL model in all cases while requiring an order of magnitude less training time.
+
+DL is a tool and, like any other tool, must be used after consideration of its strengths and weaknesses for the problem at hand.
```
SiminaB, line 50: Can start sentence with "In conclusion, DL is a tool..."

content/06.know-your-problem.md
```diff
@@ -19,13 +19,16 @@ Study designs will often have different assumptions and caveats, and these canno
 Many datasets are now passively collected or do not have a specific design, but even in this case it is important to know how individuals or samples were treated.
 Samples originating from the same study site, oversampling of ethnic groups or zip codes, and sample processing differences are all sources of variation that need to be accounted for.
 
+In all cases, investigators should consider the extent to which the outcome of interest is likely to be predictable from the input data and begin by thoroughly inspecting the input data.
+Data exploration with unsupervised learning and data visualization can reveal the biases and technical artifacts in these datasets, providing a critical first step to assessing data quality before any deep learning model is applied.
+In some cases, these analyses can identify biases from known technical artifacts or sample processing which can be corrected through preprocessing techniques to support more accurate application of deep leaning models for subsequent prediction or feature identification problems from those datasets.
```
SiminaB, line 24: I would remove "from those datasets" (I think it's implied)
"deep learning" -> "DL"

content/09.overfitting.md
```diff
@@ -8,6 +8,12 @@ To continue the student analogy, a smarter student has greater potential for mem
 
 ![A visual example of overfitting and failure to generalize. While a high-degree polynomial gets high accuracy on its training data, it performs poorly on data unlike that which it has seen before. In contrast, a simple linear regression works well on both datasets. The greater representational capacity of the polynomial is analogous to using a larger or deeper neural network.](images/overfitting.png){#fig:overfitting-fig}
 
+To evaluate deep supervised learning models, they should be trained, tuned, and tested on non-overlapping datasets.
```
SiminaB, line 11: "deep learning" -> "DL"

content/09.overfitting.md
```diff
@@ -8,6 +8,12 @@ To continue the student analogy, a smarter student has greater potential for mem
...
+The data used for testing should be locked and only used one-time for evaluating the final model after all tuning steps are completed.
+Using a test set more than once will lead to biased estimates of the generalization performance [@arxiv:1811.12808; @doi:10.1162/089976698300017197].
+Additionally, many conventional metrics for classification (e.g. area under the receiver operating characteristic curve or AUROC) have limited utility in cases of extreme class imbalance [@pmid:25738806].
+Model performance should be evaluated with a carefully picked panel of relevant metrics that make minimal assumptions about the composition of the testing data [@doi:10.1021/acs.molpharmaceut.7b00578], with particular consideration given to metrics that are most directly applicable to the task at hand.
```
SiminaB, line 15: Give specific example of relevant metrics?

content/11.interpretation.md
```diff
@@ -2,7 +2,7 @@
 
 Once we have trained an accurate deep model, we often want to use it to deduce scientific findings.
 In doing so, we need to take care to correctly interpret the model's predictions.
-We know that the basic tenets of machine learning also apply to deep learning ([Tip 1](#concepts)), but because deep models can be difficult to interpret intuitively, there is a temptation to anthropomorphize the models.
+Because deep models can be difficult to interpret intuitively, there is a temptation to anthropomorphize the models.
```
SiminaB, line 5: What does it mean to "anthropomorphize the models?" I know the meaning of the term and it's application as a literary device, but not sure of its exact meaning here...

content/09.overfitting.md
SiminaB: Not sure it's worth mentioning that if a single dataset is split into 2 or 3 for training, tuning, testing and preprocessing (like quantile normalization) is applied before this split, then the resulting datasets may not be truly independent.

SiminaB (changes requested):
Looks good in general! Some generally small comments here and there, a few of which may require a couple of extra sentences.

Benjamin-Lee:
@SiminaB thanks for the feedback. Going to bed now but will work on it tomorrow.

content/06.know-your-problem.md
cgreene: My 2 cents: I would go the other direction - i think we should just use "deep learning" as the written out text throughout.

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+An advantage of ML over DL is that currently there are more tools capable of automating the model selection and training process.
+Automated ML (AutoML) tools such as TPOT [@doi:10.1007/978-3-319-31204-0_9], which is capable of using genetic programming to optimize ML pipelines, and Turi Create [@https://github.com/apple/turicreate], a task-oriented ML and DL framework which automatically tests multiple ML models when training, allow users to achieve competitive performance with only a few lines of code.
+Luckily, there are efforts underway to reduce the expertise required to build and use DL models.
+Indeed, both TPOT and Turi Create, as well as other tools such as AutoKeras [@arxiv:1806.10282], are capable of abstracting away much of the programming required for "standard" DL tasks.
+Projects such as Keras [@https://keras.io], a high-level interface for TensorFlow, make it relatively straightforward to design and test custom DL architectures.
+In the future, projects such as these are likely to bring DL experimentation within reach to even more researchers.
+
+There are some types of problems in which using DL is strongly indicated over ML.
```
cgreene, line 36: I think this part is missing "why" to synthesize the examples.

content/11.interpretation.md
cgreene: I see this happening a lot when folks say the network learns to recognize lines, then a beak, then a bird in neural network features of increasing complexity (or similar). It's hard for me to see this paper and view that statement as consistent with these results: https://distill.pub/2017/feature-visualization/

I agree with @SiminaB we need a reference and example here.

cgreene:
Of the presented options I think I prefer "Decide whether deep learning is appropriate for your problem"

content/06.know-your-problem.md
SiminaB: Either way is fine! No strong opinion from me as long as we're being consistent.

agitter: I also prefer writing out "deep learning".  For a short paper, we can avoid most or all acronyms.

agitter:
I also like "Decide whether deep learning is appropriate for your problem".  It suggests that you have a goal (problem) and want to consider appropriate tools, which may or may not be deep learning.

Benjamin-Lee:
@cgreene and @agitter I've gone and renamed the tip. I think it sounds better also and is less negative than the other option

content/06.know-your-problem.md
Benjamin-Lee: We're already well over the editor's suggested word limit of 2500-3000 words. I personally prefer to eliminate the use of "DL" for "deep learning" for several reasons in order of strength:

1. Fewer acronyms to keep in mind for the reader
2. DL isn't a super common acronym in general. For example, when you google "ML" the first thing you get is Merrill Lynch and the next thing is a bunch of videos on TF followed by an ML cause from Stanford. For "DL", nothing deep learning even shows up.
3. I'm not entirely sure if this applies to academic papers, but search engine optimization is worth considering. For search methods which rely on term frequency of body content, using the keyword "deep learning" more will make the paper more visible than only having one or two occurrences.

Since there are at least three authors (@cgreene, @agitter, and myself) in favor of the change and none opposed, I think we should do it. I've made a new issue (#264) to track it. In the meantime, we can be inconsistent until normalization after we freeze the content (once these PRs are done).

content/03.when-to-use-dl.md
Benjamin-Lee: @beamandrew can you help out with this sentence? I see you have a blog post on the topic: https://beamandrew.github.io/deeplearning/2017/06/04/deep_learning_works.html

Benjamin-Lee: We can probably revise the number down to ~100 and cite this: https://arxiv.org/abs/1511.06348

We should also mention one-shot learning, but saying that you require a minimum of one data point in your training data is somewhat useless for the beginner deciding whether to use DL.

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+Given DL's usefulness across a range of scientific questions and data modalities, it may appear that it is a panacea for modeling problems.
+Indeed, neural networks are universal function approximators, meaning that they are in principle capable of learning any function [@doi:10.1007/BF02551274; @tag:hornik-approximation].
+If DL is so powerful and popular, why would one ever not choose to use it?
+
+The reason is simple: DL is not suited to every situation in reality.
+Training DL models requires a significant amount of data, computing power, and expertise.
+In some areas of biology where data collection is thoroughly automated, such as DNA sequencing, large amounts of quality data may be available.
+For other areas which rely on manual data collection, there may not be enough data to effectively train models.
+Though there are methods to increase the amount of training data such as data augmentation (in which existing data is slightly manipulated to yield "new" samples) and weak supervision (in which simple labeling heuristics are combined to produce noisy, probabilistic labels) [@arxiv:1605.07723v3], these methods cannot overcome a complete shortage of data.
```
ajlee21, line 12:
```suggestion
Though there are methods to increase the amount of training data, such as data augmentation (in which existing data is slightly manipulated to yield "new" samples) and weak supervision (in which simple labeling heuristics are combined to produce noisy, probabilistic labels) [@arxiv:1605.07723v3], these methods cannot overcome a complete shortage of data.
```

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+Projects such as Keras [@https://keras.io], a high-level interface for TensorFlow, make it relatively straightforward to design and test custom DL architectures.
+In the future, projects such as these are likely to bring DL experimentation within reach to even more researchers.
+
+There are some types of problems in which using DL is strongly indicated over ML.
+Assuming a sufficient quantity of quality data is available, applications such as computer vision and natural language processing are likely to benefit from DL.
+In fact, these areas were the first to see significant breakthroughs through the application of DL [@doi:10.1145/3065386] during the recent "DL revolution."
+For example, Ferreira et al. used DL to recognize individual birds from images [@doi:10.1111/2041-210X.13436].
+This problem was historically difficult but, by combining automatic data collection using RFID tags with data augmentation and transfer learning (explained in [Tip 5](#architecture), the authors were able to use DL achieve 90% accuracy in several species.
```
ajlee21, line 40:
```suggestion
This problem was historically difficult but, by combining automatic data collection using RFID tags with data augmentation and transfer learning (explained in [Tip 5](#architecture)), the authors were able to use DL achieve 90% accuracy in several species.
```

content/03.when-to-use-dl.md
```diff
@@ -0,0 +1,51 @@
...
+As an illustration, Rajkomar et al. [@doi:10.1038/s41746-018-0029-1] found that simpler baseline models achieved performance comparable with that of DL in a number of clinical prediction tasks using electronic health records, which may be a surprise to many.
+Another example is provided by Koutsoukas et al., who benchmarked several traditional machine learning approaches against deep neural networks for modeling bioactivity data on moderately sized datasets [@doi:10.1186/s13321-017-0226-y].
+The researchers found that while well tuned DL approaches generally tend to outperform conventional classifiers, simple methods such as Naive Bayes classification tend to outperform DL as the noise in the dataset increases.
+Similarly, Chen et al. [@doi:s41746-019-0122-0] tested DL and a variety of traditional ML methods such as logistic regression and random forests on five different clinical datasets, finding that the non-DL methods matched or exceeded the accuracy of the DL model in all cases while requiring an order of magnitude less training time.
+
+DL is a tool and, like any other tool, must be used after consideration of its strengths and weaknesses for the problem at hand.
```
ajlee21, line 50:
```suggestion
In conclusion, deep learning is a tool and, like any other tool, must be used after consideration of its strengths and weaknesses for the problem at hand.
```

ajlee21 (commented):
I like the new tip. I've just made a few minor inline changes based on people's comments

content/03.when-to-use-dl.md
Benjamin-Lee: Fixed. Thanks!

content/03.when-to-use-dl.md
Benjamin-Lee: Basically, what I'm getting at is that you don't need to start out with the most high powered tools there are now that there are more beginner-friendly ones that are stable. If you need a NN architecture that doesn't exist out of the box in one of the higher level tools but isn't so exotic that you _need_ TensorFlow primitives, a tool like Keras might be a good middle ground.

content/09.overfitting.md
Benjamin-Lee: @siminab this is a good point but it should go in the overfitting tip.

Benjamin-Lee: Ok I made a new issue for that

Benjamin-Lee: Fixed

content/03.when-to-use-dl.md
Benjamin-Lee: Good point. Fixed!

rasbt (approved):

content/03.when-to-use-dl.md
Benjamin-Lee: Ok I added the citation and revised it down.

content/09.overfitting.md
Benjamin-Lee: I don't want this to block merging so I've made a new issue #267 to track this since we should clarify

content/11.interpretation.md
Benjamin-Lee: Tracked in #268

ttriche:
quality exposition from Daniela Witten on how regularization/penalization
and stochastic gradient descent can be viewed as two sides of the same
coin:

https://threadreaderapp.com/thread/1292293102103748609.html

same phenomenon observed with gradient boosting converging to an L1
penalized regression fit, etc.

as Hastie has said a nearly infinite number of times, if the true model is
simple, penalization will find it; if the true model is complex, only more
data will find it.

(with the caveat that truth is subjective and not necessarily important :-)
)


--t

PR #253, opened 2020-10-09
Update 02.intro.md

mdkessler:
Suggested edits for adding context and readability for readers new to computational science.

- [x ] Yes, I added myself or am already a contributor

**Any more details?**


content/02.intro.md
```diff
@@ -1,12 +1,19 @@
...
-By increasing the accessibility of DL for applications in biological research, we aim to improve the overall quality and reporting of DL in the literature, enabling more researchers to utilize these state-of-the art modeling techniques.
+Machine learning (ML) is a branch of computational science organized around the learning by computers via algorithms of relationships between dataset variables. Traditionally this has referred to statistical methods such as linear and logistic regression, but as the scale and complexity of data have increased, many new ML methods have been developed to model and learn increasingly complex relationships.
+One powerful approach for learning complex (i.e. linear and non-linear) relationships between input and output variables is called the artificial neural network, which makes use of a number of serialized regression frameworks (i.e. layers) in order to learn such complexity (cite).
```
rasbt, line 4: usually i.e. should be followed by a comma. "i.e." -> "i.e.,"

rasbt: I don't understand the "serialized regression frameworks" -- maybe "series of regression models" but it's still very confusing to say this. Can we replace

"called the artificial neural network, which makes use of a number of serialized regression frameworks (i.e. layers)" 


with sth along the lines of

"called the artificial neural network, which makes use of a series of affine transformations accompanied by non-linear activation functions(i.e., the neural network layers)"

rasbt (changes requested):

ejfertig:
I agree with these comments. A lot of the introduction is written currently assuming a deep understanding of DL. E.g., it starts with "Deep learning (DL) is a subfield of machine learning (ML) focusing on artificial neural networks with many layers. " I thinnk the idea of both neural networks, layers, and what this gives as Michael is outlining here is important to make it understandable to a novice.

content/02.intro.md
Benjamin-Lee: I worry the change @rasbt proposed may be incomprehensible to beginners. Someone who doesn't know what a neural network is unlikely to have it clarified by describing it as a "series of affine transformations accompanied by non-linear activation functions". 

What about:

```suggestion
One powerful approach for learning complex (that is, both linear and non-linear) relationships between input and output variables is called an artificial neural network.
In these networks, which are modeled on the human brain, artificial neurons arranged into layers "fire" based on the output of the previous layer combined with a modifiable parameter, called a weight. 
Thus, the process of "training" a neural network is in fact the tuning of these weights in such a way that a metric, called a cost function, is optimized.
```

mdkessler: @Benjamin-Lee Yea, that what I was thinking too - that describing it in the introduction as a "series of affine transformations accompanied by non-linear activation functions" would likely still be somewhat inaccessible for many biologists. I like your suggested edit - I think that may be clearer for a wider audience.

mdkessler, line 4: @Benjamin-Lee Just a few minor suggestions

```suggestion
One powerful approach for learning complex (that is, both linear and non-linear) relationships between input and output variables is called an artificial neural network.
These networks are modeled after the human brain, and they comprise artificial neurons arranged into layers.
Each layer receive input from previous layers (the first of which represents the input data), and then "fires off" its own weighted output that then serves as input into subsequent layers in the network. 
Thus, the process of "training" a neural network is in fact the tuning of these output weights such that a metric called a cost function is optimized.
```

Benjamin-Lee (approved):
LGTM

Benjamin-Lee:
Sorry for spamming everyone on the review requests; just want to get @mdkessler's changes reviewed and merged to make space for @rasbt to review.

pstew (approved):

PR #255, opened 2020-10-09
Mdkessler fix tip 1

mdkessler:
Suggested edits for additional context and readability for those without computational background

**Any more details?**


Benjamin-Lee:
@mdkessler is this content the same as in #253? If so, I'll close that one and we can focus on this one here.

mdkessler:
@Benjamin-Lee No it is different. This one updated the first tip (ml-concepts). #253 has edits to the into. I'm making another PR now that can replace #253 though - it will be the same content, but broken per sentence as we discussed. I'll also add clearer comments to each PR moving forward.

Benjamin-Lee:
@mdkessler as a note, we're working on a complete rewrite of the first tip, changing its focus to when to use DL in the first place. The current content in it is probably going to be moved elsewhere so the improvements in this patch will likely be moved to different files

mdkessler:
@Benjamin-Lee Sounds good. Thanks for the heads up.

Benjamin-Lee:
@mdkessler I'm going to close this since the majority of the content you fixed ended up being deleted or thoroughly modified. The area that remains is now in the overfitting tip, so we can move that there when you reach it. Thanks again for the help!

PR #269, opened 2020-10-17
Remove simulated data from tip 4 and change tip 5 to focus on reproducibility

Benjamin-Lee:

**Any more details?**

This PR addressed two issues:

1. The data simulation part of tip 4 is awkward, both in content and style. 
2. Tip 4 had some reproducibility stuff that I thought should be in tip 5.

When addressing (2), I noticed that Tip 5 was really about reproducibility, especially since we talk about training complexity in the first tip now. I also saw this paper (https://www.nature.com/articles/s41586-020-2766-y) earlier today, which made me think we should address code sharing as distinct from data sharing.


Benjamin-Lee:
Also, I removed a paragraph which is essentially links to three different rules.

cgreene (approved):

cgreene:
Changes look good to me. I can't remember where that simulation ended up coming from initially but it has been moved all over the full manuscript. Maybe it's rule 11, which doesn't quite make a ten simple rules list :)

pstew (approved):

PR #282, opened 2020-11-10
Update rootstock on 2020-11-10

agitter:
This rootstock update patches two severe issues in the build environment. One is causing builds to fail now. The other may cause builds to start failing in a week.


cgreene (approved):

agitter:
The build logs look good except for the broken DOI reference I fixed in 503049a

Benjamin-Lee (approved):

Benjamin-Lee:
Thanks @agitter!

PR #294, opened 2020-12-03
Minor text edit and tried to fix citation tags

ajlee21:

- [ ] Yes, I added myself or am already a contributor

**Any more details?**

Reading through the manuscript it looks great! Found 2 citations that weren't recognized and so tried to fix these (though I'm not sure if the fixes worked because I wasn't sure how to see the updated compiled manuscript). I also made one minor text edit.


Benjamin-Lee:
I seem to recall the Hornik DOI not working due to it having parentheses. If the build passes, it should be GTG

ajlee21:
> I seem to recall the Hornik DOI not working due to it having parentheses. If the build passes, it should be GTG

I see. I guess if that's the case I'm not sure what the issue with that tag is.

Benjamin-Lee (approved):
Looking at the rendered PDF, its GTG

rasbt (approved):
LGTM

PR #295, opened 2020-12-03
Fix DOI references

agitter:

**Any more details?**
Follow up to #294.  The HTML version on GitHub Pages is having problems with these references.  I'm getting rid of the `hornik-approximation` tag everywhere else so that there isn't confusion about whether to use the tag or short DOI.


ajlee21 (approved):
LGTM..finally found where the built manuscript file was and it looks fixed!

Benjamin-Lee (approved):

Issue #261, opened 2020-10-11
Add a figure/infographic listing the ten tips?


Benjamin-Lee:
I was just looking online and saw someone tweeted a like to a Ten Simple Rules article which featured a figure showing the ten rules. On Twitter, it looked like this:

![image](https://user-images.githubusercontent.com/13618205/95688479-6e854000-0bd8-11eb-868f-4403af78b4bb.png)

Does anyone think this is a good idea for outreach?

cgreene:
Yes - it also makes us keep the rules simple! 😁

Benjamin-Lee:
Ok we can work on this after we freeze the content and start soliciting final approvals.

rasbt:
Yeah, I think it would be nice to have a small infographic for the 10 tips like the one above (or the one below; in nice of course)

![Screen Shot 2020-10-12 at 9 10 56 PM](https://user-images.githubusercontent.com/5618407/95807334-868bbb00-0ccf-11eb-9521-985a8ec92a65.png)

But I'd say we should wait until any sharing/advertising until after the peer-reviews and potential acceptance for publication.

Benjamin-Lee:
@rasbt what are your thoughts on including the infographic as fig 1 in the paper? That's what the paper I saw on twitter did (and why that was the thumbnail). As it stands, right now our thumbnail is the overfitting figure.

rasbt:
Oh this was in the paper itself? I see. Hm, it's basically just the section titles, right? To be honest, having a figure listing the section titles sounds a bit overkill since the figure doesn't really add anything to the paper then. We could maybe have it as a summary in the beginning and end if it has some additional keyword phrases that summarize each section.

Benjamin-Lee:
Yeah they put it in the introduction: https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1008210

I haven't seen that done before which is why I raised it in an issue.

rasbt:
Hm, I think for our paper this would be a bit too much -- we already have figures, so there is no need to create figures for the sake of having figures in the paper. However, we should probably make a flowchart figure for sharing it later on if/when it is accepted & published

chevrm:
This one is a pretty nice aesthetic
https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1007809

rasbt:
I agree. This one, with its short summaries, looks nice & useful to me

fmaguire:
[...]

PR #298, opened 2020-12-18
Removes cross-links given shortness of document

### LINKED ISSUE #266: Remove links to tips
There are quite a few tips with these cross references. As I'm reading I do find them a bit odd for this type for a manuscript that is not _that_ long. I'd probably remove them throughout, unless someone has a strong case for keeping them.

_Originally posted by @cgreene in https://github.com/Benjamin-Lee/deep-rules/pull/258#discussion_r503165828_

fmaguire:
Closes #266

And provides a draft figure for #261


rasbt (changes requested):
The figure is really awesome!

Below are just some suggested tweaks to the text:

Since we use AE in the rest of the text, changing the following would be good for consistency:

6) optimisiation -> optimization
7) regularise -> regularize

We also decided to remove 2nd person everywhere. So, below are some suggestion (but there may be better ways to reword those instances)

2) your DL model -> a DL model
4) peculiarities of your data -> peculiarities of the data
8) why your model -> why a model

Lastly, maybe the following change would make it sound a tad letter in 9)

9) derived from model -> derived from a model

Benjamin-Lee:
Is it standard to have a figure before the introduction heading? Or should it go immediately after?

Benjamin-Lee:
Also, this figure is 100% what we should be using as a thumbnail: https://github.com/Benjamin-Lee/deep-rules/blob/master/thumbnail.png

Can you update that file as well?

rasbt:
> Is it standard to have a figure before the introduction heading? Or should it go immediately after?

Some papers have so-called "graphical abstracts" which are essentially figures that go next/below the abstract. Otherwise it is not standard. My Ph.D. advisor was pretty strict about that: she never allowed a figure to appear before it is referenced in the text. I think most people are more relaxed about that these days but it would maybe be weird to put it above the heading if it is not a graphical abstract.

fmaguire:
A graphical abstract was kinda my thinking but happy to move it into the introduction.

I'm knee deep in the latest covid variant stuff but will try to get that updated and thumbnail added ASAP

Benjamin-Lee:
No rush at all. Thanks for the great work!

rasbt:
> A graphical abstract was kinda my thinking but happy to move it into the introduction.

I think that would be great, we just have to make sure whether the journal supports that, @Benjamin-Lee do you know more about this?

Benjamin-Lee:
I've never seen it at PLOS Comp Bio, only figure 1 pseudo-graphical abstracts. This is their [page on figures](https://journals.plos.org/ploscompbiol/s/figures) but it clearly either doesn't apply to Ten Simple Rules articles or isn't enforced. Specifically, it says not to use article title (violated in the antiracist lab paper) and to use Times/Helvetica font (violated by PE Bourne himself).

fmaguire:
Updated as per issue discussion.

content/02.intro.md
```diff
@@ -21,6 +22,8 @@ To make deep learning more accessible to biological researchers, we solicited in
 These individuals collaboratively contributed to this manuscript's writing using the GitHub version control platform [@url:https://github.com/Benjamin-Lee/deep-rules] and the Manubot manuscript generation toolset [@url:https://greenelab.github.io/meta-review/].
 The goal was to articulate a practical, accessible, and concise set of guidelines and suggestions for biologically oriented researchers to follow when using deep learning.
```
agitter, line 23:
```suggestion
The goal was to articulate a practical, accessible, and concise set of guidelines and suggestions for biologically oriented researchers to follow when using deep learning (Figure @fig:overview-fig).
```

This figure placement looks good to me, but I think we need to reference it in the text as well.

fmaguire: Good shout, done.

rasbt (approved):

agitter (approved):
It looks great in the PDF artifact

PR #310, opened 2021-01-13
Remove question mark from tip 1 overview figure

fmaguire:
Accidentally kept a question mark in the first tip of the full infographic

Addresses @juancarmona 's comment in #306


Benjamin-Lee:
Thanks! In #305 there were four files changed bit this modifies two. Is that intentional?

fmaguire:
Yup, it was fine in the thumbnail version so that didn't need changed, only the full figure svg and png are updated.

Benjamin-Lee:
Got it. Merging now!

PR #316, opened 2021-01-22
Update overfitting figure

### LINKED ISSUE #280: Get rid of overfitting figure?
This figure has been a point of contention for a while. @SiminaB raised this concern (https://github.com/Benjamin-Lee/deep-rules/pull/275#discussion_r508919463):

> I'm not totally sure about this figure, just because even for linear regression, it's not recommended to use the predictions outside the range.

I seem to recall other people having various issues with it. I see several options:

1. Keep it as is
2. Add additional explanation to the figure caption
3. Change the figure
4. Eliminate the figure

I'm right now leaning towards eliminating it even though it was a good amount of work (trying various random seeds, polynomial degrees, etc) to make. Some reasons I think this is the best course of action are as follows:

1. It isn't actually an example of overfitting but rather a failure to generalize to out of range data. For DL models this distinction isn't always as clear cut (does showing x-rays from one hospital to a model trained at another count as out of range?) but for the linear regression it is.
2. There's nothing particularly novel about our figure showing overfitting. @rasbt has one in his excellent book and I'd rather point readers there.
3. If we do end up going with Nature Computational Science, xkcd style won't cut it.
4. It will possibly cause us issues down the line in review if we aren't very careful about how we describe the figure.

The main current upside is that it is only visual example we have in our paper. For visual people, figures/infographics can be quite helpful. 

If anyone is in favor of the figure, please do speak up.

agitter:

**Any more details?**
Closes #280

This update moves the test points to lie within the same domain as the training points.  I updated the caption as well.  I can move these points if anyone has suggestions.

I also updated the caption and added a reference about overparameterized neural networks.  I don't want to get into the double descent phenomenon, but it isn't accurate to say that more parameters leads to overfitting in deep learning.


content/09.overfitting.md
```diff
@@ -6,8 +6,9 @@ In other words, the model fits patterns that are overly specific to the data it
...
+However, having a large number of parameters does not always imply that a neural network will overfit [@doi:10.1073/pnas.1903070116].
 
-![A visual example of overfitting and failure to generalize. While a high-degree polynomial achieves high accuracy on its training data, it performs poorly on data with specificities that have not been seen before. That is, the model has learned the training dataset specifically rather than learning a generalizable pattern that represents data of this type. In contrast, a simple linear regression works well on both datasets. The greater representational capacity of the polynomial is analogous to using a larger or deeper neural network.](images/overfitting.png){#fig:overfitting-fig}
+![A visual example of overfitting and failure to generalize. While a high-degree polynomial achieves high accuracy on its training data, it performs poorly on the test data that have not been seen before. That is, the model has memorized the training dataset specifically rather than learning a generalizable pattern that represents data of this type. In contrast, a simple linear regression works well on both datasets.](images/overfitting.png){#fig:overfitting-fig}
```
rasbt, line 11: >  In contrast, a simple linear regression works well on both datasets.

"well" is relative. The overfitting model performs well on the training set and not well on the test set. The performance of the linear model on both datasets falls between the aforementioned training and test performance. 

Maybe we should change

> "regression works equally well on both datasets."

instead of 

> regression works well on both datasets.

I.e., focusing more on the different between training and test performance. 

I don't feel strongly about that though, because I think it should be obvious what the figure aims to convey.

rasbt:
Besides my minor comment above, the change looks nice!

agitter:
> "well" is relative. The overfitting model performs well on the training set and not well on the test set. The performance of the linear model on both datasets falls between the aforementioned training and test performance.

I agree with this.  Since I'm remaking the figure anyway, should I decrease the error?  The point may be more natural if both the training and test data have a cleaner linear fit.

@rasbt do you have any thoughts on overparameterization and how overfitting with neural networks can be counter-intuitive?

rasbt:
> I agree with this. Since I'm remaking the figure anyway, should I decrease the error? The point may be more natural if both the training and test data have a cleaner linear fit.

Yes, that may be a good idea.

> @rasbt do you have any thoughts on overparameterization and how overfitting with neural networks can be counter-intuitive?

Hm, I am not exactly sure what you have in mind. One weirdness/counterintuition that comes to mind is the double-descent phenomenon (https://arxiv.org/abs/1912.02292), but maybe that's out of the scope for this article?

SiminaB (approved):
Looks good to me!

agitter:
Yes, double descent is what I had in mind.  I added
> However, having a large number of parameters does not always imply that a neural network will overfit [@doi:10.1073/pnas.1903070116].

with reference to https://doi.org/10.1073/pnas.1903070116 for more information.  I don't think we have space to discuss double descent and related theory, but I don't want to give the wrong impression that more parameters in a neural network always increases the chance of overfitting.

I rephrased the caption per your suggestion.

Here are more alternative versions of the figure that have less noise.  The first and last are my favorites because the polynomial stays within the y range.  The first may be slightly better because it has less noise and still makes the point.

delta 1.0, seed 3:  image

delta 1.0, seed 10: image

delta 1.5, seed 3: image

delta 1.5, seed 10: image

rasbt:
> with reference to https://doi.org/10.1073/pnas.1903070116 for more information. I don't think we have space to discuss double descent and related theory, but I don't want to give the wrong impression that more parameters in a neural network always increases the chance of overfitting.

I agree, I wouldn't dive into too much detail here, but adding it for technical completeness/correctness makes sense. The way you referenced it looks good and sufficient to me!

The figures look all nice. I have a slight preference for the first one but the other ones are great too

rasbt (approved):

rasbt:
Great job in this PR. Glad to seeing the paper getting closer to submission now :)

PR #324, opened 2021-02-02
Add zipcode metadata required for submission

rasbt:

Adds zip code to author data

### DIFF
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -79,7 +79,7 @@ authors:
     affiliations:
       - Department of Statistics, University of Wisconsin-Madison
     zipcode:
-      - 53706
+      - "53706"
   -
     github: fmaguire
     name: Finlay Maguire
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -78,6 +78,8 @@ authors:
     conflicts: None
     affiliations:
       - Department of Statistics, University of Wisconsin-Madison
+    zipcode:
+      - 53706
   -
     github: fmaguire
     name: Finlay Maguire
--- a/content/metadata.yaml
+++ b/content/metadata.yaml
@@ -79,6 +79,8 @@ authors:
     affiliations:
       - Department of Statistics, University of Wisconsin-Madison
     postcode: "53706"
+    approval: yes
+    accountable: yes
   -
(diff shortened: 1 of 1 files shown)


Benjamin-Lee (approved):

Benjamin-Lee:
Actually, does it make sense to have the value be a list of the number itself? And speaking of which, should we represent it as a string? I do know horrible things can happen when you treat zip codes as numbers, though I doubt it will matter here.

rasbt:
> Actually, does it make sense to have the value be a list of the number itself?

I was thinking that if affiliations is a list, then the zip code should probably also be a list. Although we really don't need multiple zip codes for the submission.

>  I do know horrible things can happen when you treat zip codes as numbers, though I doubt it will matter here.

Yeah, I don't think it matters here. I was thinking of it as just something you would copy and paste into the submission form.

cgreene:
What about just asking that folks include zipcodes for at least one affiliation on that line? I'm wondering if that's easier or harder.

fmaguire:
Also worth noting that not all currently listed authors are based in the US (including myself), therefore don't have zipcodes.

rasbt:
Either way would be fine to me. I think @Benjamin-Lee would just copy & paste the first value into the submission field anyway. I can change it to a single value.

Benjamin-Lee:
Single value works for me. Let’s just go with that since it reduces the complexity.

rasbt:
Good point by @fmaguire 

> Also worth noting that not all currently listed authors are based in the US (including myself), therefore don't have zipcodes.

for those cased, people could add a "- postalcode: ... " entry? Looking at your submission form, Ben, I think either zip code or postal code would be ok.
106531764-dda2df00-64bc-11eb-8a81-0550449bc3e0

Benjamin-Lee:
Let’s go with “postcode” then. An arbitrary string I can put in the form to make them happy will suffice.

rasbt:
> Let’s go with “postcode” then. An arbitrary string I can put in the form to make them happy will suffice.

alright, done. And like you said, i think we can be flexible because it's just something you have to copy and paste and doesn't go through computational pipelines

agitter (approved):

PR #339, opened 2021-02-17
Update contributors.md

khyu:

**Any more details?**


Benjamin-Lee:
Not needed but much appreciated! We'll probably get rid of this column since we're relying on the yaml for this.

PR #362, opened 2021-05-14
Build DOCX manuscript for journal submission

agitter:
We may need a DOCX version of the manuscript for the journal submission.  This will have GitHub Actions export a DOCX with every build.

### DIFF
--- a/.github/workflows/manubot.yaml
+++ b/.github/workflows/manubot.yaml
@@ -13,6 +13,7 @@ jobs:
     env:
       GITHUB_PULL_REQUEST_SHA: ${{ github.event.pull_request.head.sha }}
       SPELLCHECK: true
+      BUILD_DOCX: true
     steps:
       - name: Set Environment Variables
         run: |


rasbt:
I see the manubot build check was canceled, but overall this change and PR look ok

rasbt (approved):

agitter:
> I see the manubot build check was canceled

I checked GitHub status after you noted this, and GitHub Actions was having system-wide problems.  It looks like everything is working now.

Here is the DOCX file from the GitHub Actions artifact: [manuscript.docx](https://github.com/Benjamin-Lee/deep-rules/files/6481335/manuscript.docx)

It looks good to me.  The broken SVG icons in the author list are a known issue and have to be manually deleted.  Line numbers will also have to be added manually.

rasbt:
Great, seems to work now!

Benjamin-Lee:
Thanks! This was one of the things that was blocking me!

Benjamin-Lee (approved):

agitter:
If anyone is looking for the DOCX output, it will be in https://github.com/Benjamin-Lee/deep-rules/tree/output as `manuscript.docx`

PR #364, opened 2021-05-25
Switch to PLOS CSL

### LINKED ISSUE #363: Change CSL for Plos Comp Bio
I'm within an inch of submitting the paper and just realized that we have to use their reference format. I changed it in metadata.yaml according to the docs but no luck in the resulting output. @agitter or @dhimmel do you have any idea what I should do differently?

agitter:
Closes #363
Follows up on 69953e90911d9adf1735e666ee97e931eba3b46c, which added the CSL file.

The CLS file is specified in the `metadata` section of `build/pandoc/defaults/common.yaml` as opposed to the `content/metadata.yaml` file.  Sorry this was confusing.

Let's check the build before merging in case I made a mistake.


Benjamin-Lee:
Yep, the build is right. Ready to submit!

Benjamin-Lee:
Thanks @agitter for figuring it out

PR #370, opened 2021-11-15
added study supporting ML as baseline

rasbt:
Saw a study that supports one of our claims re ML baselines and added it


agitter (approved):
I agree we should support this statement.  This paper is a good addition.  It focuses mostly on computing time and costs, so we may want another 1-2 papers about performance of simpler models versus deep learning.

rasbt:
I agree. It's very hard to find these papers though. The problem is usually that papers that involve deep learning are centered around deep learning, and the authors try to show that it outperforms other methods. 

One recent paper that comes to mind was related to tabular datasets, i.e., here one take-away was that DL works well on tabular data, but non-DL methods (here: gradient boosting) can perform similarly well: https://arxiv.org/abs/2106.11959. The one problem here is though that gradient boosting is not necessarily simpler than DL

Benjamin-Lee (approved):

PR #372, opened 2021-11-27
Additional edits in response to reviews

agitter:
I made edits to our revisions, updated references, and did some additional proofreading.

@rasbt I moved one of the references you added (https://dl.acm.org/doi/abs/10.1145/3106237.3106256).  It seemed to fit better in the paragraph where we talk about advantages of traditional ML over DL.  I added a reference to https://www.nature.com/articles/s41580-021-00407-0 in its place and also found this study about linear models generalizing better than autoencoders (https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-020-3427-8).


rasbt (approved):
Thanks for fixing those!

